PR #72, opened 2014-12-21
Entities for build instructions fixes #70

### LINKED ISSUE #70: How do we want to bind build instructions to courses or groups?
Currently the maven build instruction is hard coded for the git hook. When we want to allow build instructions to be different for courses or even specific groups, we want to store this in Devhub. But how should we do this?

Some of the possibilities that I see:
* Create new entities in Devhub that can create a `BuildInstruction`. A `Group` or `Course` will then connect to this.
* Add JPA annotations to the existing `BuildInstructions` in the build models repository. This requires an additional `@Id` field for the `BuildInstruction` and introduces a `javax.persistance` dependency
* Store the serialized form of `BuildInstruction` (either JSON or through `Serializable`)

JWGmeligMeyling:
Proposal for string build instructions in Devhub (#70)


src/main/resources/META-INF/persistence.xml
```diff
@@ -6,6 +6,15 @@
 
 	<persistence-unit name="default" transaction-type="RESOURCE_LOCAL">
 		<provider>org.hibernate.ejb.HibernatePersistence</provider>
+		<class>nl.tudelft.ewi.devhub.server.database.entities.BuildResult</class>
```
michaeldejong: There shouldn't be a need to manually register classes in the current setup. Have you tested without this?

michaeldejong:
Check my comments, and then rebase on master. :smile:

src/main/resources/META-INF/persistence.xml
JWGmeligMeyling: It works without, but what is better/worse with registering this?

michaeldejong: If you ever want to add, remove, or rename entities you have to update this file. Hence autodiscover is easier and less error-prone for devs.

PR #73, opened 2014-12-22
Added tests for entities and controllers, fixed some found issues

JWGmeligMeyling:
This PR comes with some tests for the controller classes and small fixes for problems found in testing.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Course.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Course.java
@@ -39,7 +39,7 @@ public class Course {
 	private String name;
 
 	@NotEmpty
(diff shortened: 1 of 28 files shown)


src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Course.java
```diff
@@ -70,6 +70,10 @@
 	@OneToMany(mappedBy = "course", fetch = FetchType.LAZY)
 	private List<CourseAssistant> courseAssistants;
 
+	public void setCode(String code) {
+		this.code = code.toUpperCase();
```
JWGmeligMeyling: Now enforcing unique course code (**case insensitive**, as the repository names are care insensitive).

PR #77, opened 2015-01-22
Merge branch for PR's  #68 #69 #73 #76 #78 #79 & #80

JWGmeligMeyling:
Merge branch for PR's  #68 #69 #73 #76 #78 #79 and #80


pom.xml
```diff
@@ -252,6 +252,12 @@
 			<version>2.43.1</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.jukito</groupId>
```
michaeldejong: Cool, never heard of this one before!

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Group.java
```diff
@@ -39,7 +39,7 @@
 
 	@NotNull
 	@Column(name = "group_number")
-	private long groupNumber;
+	private Long groupNumber;
```
michaeldejong: Is this nullable now?

michaeldejong:
:+1:

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Group.java
JWGmeligMeyling: No, there is a test case for this I believe. It's enforced in the database.

michaeldejong: If it's not nullable, we should keep this on `long` instead of `Long`. Saves us from auto-boxing and potential NPEs (and checks).

JWGmeligMeyling: I changed this to a `Long` because a primitive defaults to `0` and I'm not sure if that is desired behavior

michaeldejong: But if it's null you cannot persist it anyway. If NULL is not a valid value, you should use primitives if possible.

JWGmeligMeyling: Makes sense, I will change this!

PR #119, opened 2015-04-21
Fixed adding assistants with ucase

JWGmeligMeyling:
For me this fixed the crashes when adding an assistant with capital characters instead of lowercase characters. Not urgent because after cleaning your cookies you can reset the assistants again.


JWGmeligMeyling:
Same problem for groups, adding one more patch

PR #123, opened 2015-04-21
Fix for broken download link binaries in file explorer

JWGmeligMeyling:
I accidentally broke the download link for binary files in the file explorer.


JWGmeligMeyling:
Ah je merged ze nu snel, is deze weg gevallen? @michaeldejong

PR #133, opened 2015-04-22
Extracted valid SSH-Key name

JeremybellEU:
Extracted the regex match for a valid SSH-key, since this was used on multiple places. This increases maintainability in the event of editing this regex which must effect other parts of this class too.


JWGmeligMeyling:
Maybe squash te PR together as the last commit is really small? (Squashing is generally considered a bad practise but I think for pulls it's OK because no one really builds on top of them anyway).

There are two ways:
- Rebase on the master: `git rebase -i master`. Changing the `pick` commits to `squash` will squash them in to the above commit. For this, the first commit should stay on `pick`.
- Reset soft to master. This sets the branch pointer to the master, but you keep the working directiry. Then you can readd and make a new commit.

Both ways will create a new commit which overrides the current history of the branch, so you will have to force push. (`git push -f`)

JWGmeligMeyling:
Merged in #136

PR #134, opened 2015-04-22
Fix bug with ssh key setup access violation

JeremybellEU:
- Added a private method that can verify you can access the account page.
- Added the verification check in 'showNewSshKeyPage' as this was missing.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AccountResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AccountResource.java
@@ -53,6 +53,14 @@ public class AccountResource extends Resource {
 		this.currentUser = currentUser;
 		this.users = users;
 	}
+	
+	private void verifyCanAccessPage(String netId) throws UnauthorizedException {
+		User account = users.findByNetId(netId);
+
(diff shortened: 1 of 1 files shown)


src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AccountResource.java
```diff
@@ -53,6 +53,16 @@
 		this.currentUser = currentUser;
 		this.users = users;
 	}
+	
+	private User verifyCanAccessPage(String netId) throws UnauthorizedException {
```
JWGmeligMeyling: Maybe we want to rename this method because it also provides the user to edit (admins can edit students). But I can't think of any alternative.

JWGmeligMeyling:
Looks good :+1:

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AccountResource.java
JeremybellEU: I was thinking of it but I also couldn't think of a better name.

JWGmeligMeyling:
Merged!

PR #154, opened 2015-05-03
Initial Checkstyle configuration.

JeremybellEU:
To have a project-wide code-style, a Checkstyle configuration file can be imported into your IDE. This way all developers contributing to a repository will create code with a uniform style.

This Checkstyle configuration is a proposal and I think it is best to have a discussion first of which rules should be enabled and which shouldn't before this gets merged.


pom.xml
```diff
@@ -31,52 +32,52 @@
...
-            <releases>
-                <enabled>true</enabled>
-                <updatePolicy>always</updatePolicy>
-            </releases>
-            <snapshots>
-                <enabled>false</enabled>
-            </snapshots>
-        </repository>
-    </repositories>
+	<repositories>
```
JWGmeligMeyling: While consistent idententation is desirable, it's more important to find which commit introduced a single line when you want to find which commit introduced a problem. Therefore, I think, indentation should only change if that part of the code is changed.

pom.xml
```diff
@@ -465,6 +466,16 @@
...
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-checkstyle-plugin</artifactId>
+				<version>${checkstyle.version}</version>
+				<configuration>
+					<failOnViolation>false</failOnViolation>
+					<configLocation>configuration/checkstyle.xml</configLocation>
+					<suppressionsLocation>configuration/checkstyle-suppressions.xml</suppressionsLocation>
```
JWGmeligMeyling: Can we put this in ${builddir}? Otherwise we would have to exclude them from the built jar.

src/main/resources/configuration/moodcat.importorder
```diff
@@ -0,0 +1,6 @@
+#Organize Import Order
+#Tue Apr 28 12:34:57 CEST 2015
+3=com
```
JWGmeligMeyling: Maybe this file can be omitted?

JWGmeligMeyling:
I also have some checkstyle configuration fixes here https://github.com/devhub-tud/devhub/pull/97
But I believe that branch is quite behind master.

pom.xml
JeremybellEU: This was probably a side effect of indenting the checkstyle-plugin.

Spaces vs tabs should be a seperate issue I think as it will effect the whole code. When I initially enabled this Checkstyle rule, 6000 CheckStyle warnings were generated.

pom.xml
JeremybellEU: I think it would be better then to place this is in the reporting phase.

src/main/resources/configuration/moodcat.importorder
JeremybellEU: Yes it could be.

JeremybellEU:
Shall I just close this? Seems too big of a hassle integrating now.

PR #170, opened 2015-05-12
Added more mocked data to bootstrapper.

JeremybellEU:
- Bound GitServerClientMock to GitServerClient in order to load repositories without any hassle
- Updated Bootstrapper to also insert deliveries for groups.


src/test/java/nl/tudelft/ewi/devhub/server/backend/Bootstrapper.java
```diff
@@ -146,6 +151,9 @@ public void prepare(String path) throws IOException, ApiError, GitClientExceptio
                 assignmentEntity.setAssignmentId(assignment.getId());
                 assignments.persist(assignmentEntity);
                 log.debug("Persistted assignment {} in {}", assignmentEntity.getName(), course.getCode());
+                
+                // Store for later use to insert deliveries
+                lastAssignment = assignmentEntity;
```
JWGmeligMeyling: Could you store the assignments in a `Map<Long, Assignment>` and retrieve the assignments from there. You'll see where this comes in handy in the next comment..

src/test/java/nl/tudelft/ewi/devhub/server/backend/Bootstrapper.java
```diff
@@ -205,6 +213,16 @@ public void prepare(String path) throws IOException, ApiError, GitClientExceptio
 				catch (Exception e) {
 					projects.provisionRepository(groupEntity, members);
 				}
+				
+				Delivery delivery = new Delivery();
```
JWGmeligMeyling: Technically this is a good way to instantiate a delivery, but it's a bit inconsistent with the idea of the bootstrapper. The bootstrapper loads a configuration from a JSON file. The database should have only the data from that configuration, in order to create an reproducible test environment. This way we cannot test environments without deliveries for assignments.

Please add a new JSON entity:
```java
   @Data
    static class BDelivery {
        private Long id;
        private String name;
        private String username;
    }
```

Maybe store them in `BAssignment`:
```java
   @Data
    static class BAssignment {
        private Long id;
        private String name;
        private List<Delivery> deliveries;
    }
```

Then loop through the deliveries for the assignment, find the correct group and user (maybe also instantiate the date from a long timestamp, which can be useful for testing deadlines), and then persist the delivery.

JWGmeligMeyling:
PR looks good, but I have a few suggestions :smile:

JeremybellEU:
This pull request will be merged with #171

PR #171, opened 2015-05-12
State subselections fixes #114 

### LINKED ISSUE #114: Subcategories based on assignment status
On the assignment submission overview page (/courses/{coursId}/assignments/{assignmentId}) make the status fields under the progress bar (from #81 ) clickable. When you click on a status field, the submissions in the table below are filtered to only show the submissions with the specified status.

This way course managers/assistants can more easily track down student groups who are running behind or have trouble with a course, or during grading quickly get all ungraded (submitted) submissions.

JeremybellEU:
Added subselections of delivery states to assignment overview.


src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
```diff
@@ -94,6 +93,32 @@ public boolean lastDeliveryIsApprovedOrDisapproved(Assignment assignment, Group
...
+    /**
+     * Get the most recent delivery for every group in this assignment with the specific state
+     * @param assignment current assignment
+     * @return a list of deliveries
+     */
+    @Transactional
+    public List<Delivery> getLastDeliveries(Assignment assignment, String state) {
+    	BooleanExpression expression = QDelivery.delivery.assignment.eq(assignment);
+
+    	if (state != null) {
```
JeremybellEU: Personally not really happy with this if, but creating a state pattern just for this didn't sound right either.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Delivery.java
```diff
@@ -44,6 +44,14 @@
         public String getTranslationKey() {
             return translationKey;
         }
+        
+        /**
+         * Get the lowercase name for freemarker.
+         * @return {@link Enum#name() name()}{@link String#toLowerCase() .toLowerCase()}
+         */
+        public String lowerCaseName() {
```
JeremybellEU: I found out FreeMarker didn't parse chaining of methods, so I created this to make it look more natural than ALL-CAPS.

src/test/java/nl/tudelft/ewi/devhub/server/backend/Bootstrapper.java
```diff
@@ -205,6 +214,20 @@ public void prepare(String path) throws IOException, ApiError, GitClientExceptio
 				catch (Exception e) {
 					projects.provisionRepository(groupEntity, members);
 				}
+				
+				Delivery delivery = new Delivery();
```
JeremybellEU: As to #170 this can be improved, but I based this branch on that one to get easier testing. Might have to rebase later to fix merge issues.

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
```diff
@@ -74,15 +78,10 @@ public boolean lastDeliveryIsApprovedOrDisapproved(Assignment assignment, Group
...
     }
 
-    /**
-     * Get the most recent delivery for every group in this assignment
-     * @param assignment current assignment
-     * @return a list of deliveries
-     */
     @Transactional
-    public List<Delivery> getLastDeliveries(Assignment assignment) {
+    private List<Delivery> getLastDeliveriesList(Assignment assignment, Predicate expression) {
```
JWGmeligMeyling: Please don't break existing API's, they could be used in PR's or Freemarker.  If we think they are unused, we should annotate them and log a warning if used before removing them.

src/test/java/nl/tudelft/ewi/devhub/server/backend/Bootstrapper.java
JWGmeligMeyling: I personally like @LiamClark 's way of doing this more. See the mapper at https://github.com/LiamClark/devhub/commit/2e92648a2749cc2384a43a49290b5751b786769b . Just fetch them all, and group. We're talking about +/- 10 entities anyway.

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
```diff
@@ -94,6 +93,32 @@ public boolean lastDeliveryIsApprovedOrDisapproved(Assignment assignment, Group
...
+    	return this.getLastDeliveries(assignment, null);
+    }
+    
+    /**
+     * Get the most recent delivery for every group in this assignment with the specific state
+     * @param assignment current assignment
+     * @return a list of deliveries
+     */
+    @Transactional
+    public List<Delivery> getLastDeliveries(Assignment assignment, String state) {
```
JWGmeligMeyling: The enum type should be used for type safety.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Delivery.java
JWGmeligMeyling: If you want to render the *name* you should use it's translation key and run that through `i18n.translate(state.translationKey)`.

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
JWGmeligMeyling: I personally like @LiamClark 's way of doing this more. See the mapper at https://github.com/LiamClark/devhub/commit/2e92648a2749cc2384a43a49290b5751b786769b . Just fetch them all, and group. We're talking about +/- 10 entities anyway.

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
JeremybellEU: It does not break existing APIs, as the original getLastDeliveries function still exists. It is merely using this method as helper method. E.g. the content of the method is changed to a private helper method and only the selection of deliveries is delegated to the original API and a new call.

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
JeremybellEU: I have tried to use Delivery.State, but couldn't get the expression below working. Maybe I missed something.

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
JWGmeligMeyling: Ah then I maybe reather see it duplicated :smile: It's not really logic anyway, they're SQL builders.

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
JWGmeligMeyling: Will have a look.

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
JeremybellEU: I figured code duplication would be worse than a simple if, but yes it is debatable. Your call in the end ;p

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Delivery.java
JeremybellEU: All right I didn't know that, I will fix it =]

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Deliveries.java
JeremybellEU: But this would return all the submissions, but then in a map? Personally I think a 'and' construct is better than a map of all deliveries sorted, but I can understand the reasoning for choosing the latter. If you want this, I can change it of course.

JWGmeligMeyling:
This PR was kinda heavy weight imo. In PR #207 I have merged the bootstrapper and tooltips with an easier approach for the delivery filter.

PR #172, opened 2015-05-12
Moved to Bower for web dependencies

JWGmeligMeyling:
This PR switches to Bower for installing the web dependencies. It's not ment to go live yet, but rather investigate what switching to Bower would look like. Especially for when we want to switch to Angular later on.

The changes in this repository include:
* Removed the `jQuery`, `highlightjs`, `bootstrap` and `octicons` dependencies from the framework (saves about 8k lines)
* Added these as bower dependencies
* Created a `vendor/` folder under `src/main/resources/statics/` (gitignored and configured in `.bower_rc`)
* Changed some paths in the Freemarker templates / macro's and less file to match the new path
* Added a `bower install` call to the Maven `generate-sources` phase.

During test deployment, I found out that `bower install` through `mvn install` should not be ran as root. We have to keep this in mind while deploying the real servers.


JeremybellEU:
I have succesfully compiled and ran the tests, however it required some additional configuration on Windows.

You have to make sure that both `nodejs` and `git` are added to the `PATH` environment variable. You can do this by copy-pasting the folder-location of the respective .exes to the `PATH` variable.

Additionally for Eclipse, the pom.xml might generate a warning. However this warning does not prohibit you from succesfully compiling the code and running tests.

JWGmeligMeyling:
Replaced with #245

PR #237, opened 2015-06-18
Alignment of fold/unfold button, fixes #230

### LINKED ISSUE #230: "Fold" button misaligned when clicked
At first the "Fold" button is aligned correctly when viewing commits. However once this button is clicked it aligns incorrectly.

devhub-fold-alignment

aaronang:


JWGmeligMeyling:
:+1: Will test & merge later today. Thanks for the fix!

aaronang:
Is there a guide on how to setup a test environment? Because I fixed it, but I couldn't test it :smile:

JWGmeligMeyling:
This is a difficult one to test, because testing a diff requires a diff and this is one of the most recursive structures ever... So mocking is.. a pain.

What you can do is run the `DevhubInMockedEnvironment` class. This uses an embedded `H2` database rather than a database server, and that temporary database is filled by the `Bootstrapper`, which loads the configuration from a json file. This way you can easily setup groups/courses/assignments. But not really repositories.

The `DevhubInMockedEnvironment` can both use a real git server (`bind(GitServerClient.class).toInstance(new GitServerClientImpl(...));`) or a mocked git server: `bind(GitServerClient.class).toInstance(new MockedGitServerClient());` (Not really sure about that class name but it's something like that). You can look at the test cases how the `MockedGitServerClient` can be used, but it's quite difficult to set it up nicely.

When you want to use it against a real git server: clone the git server from the git server repo, install vagrant, run `vagrant up` to start a Gitolite server. Then start the git server.

If you want to write a unit test instead of manual testing, I would recommend to use the existing commit diff test case, and extend it with an option to fold and unfold the box.

By the way, I posted some issues under the new milestone https://github.com/devhub-tud/devhub/milestones/3.1.2 . If you're interested to look into some of these issues, let me know :smile:

aaronang:
I haven't set up a development environment yet for Devhub, so I will focus on that first. Meanwhile I will write a step by step guide on how to setup this development environment. I think it makes sense to have a guide for setting up a development environment?

PR #260, opened 2016-02-01
Enabled private repositories, fixes #252

### LINKED ISSUE #252: Support private repositories
It would be nice to be able to use Devhub for private repositories (not bound to a course)

JWGmeligMeyling:
Fixes #252


LiamClark:
we will never pay github again! +1

PR #282, opened 2016-04-16
Delivery view not rendering properly if no  commit is attached to delivery

JWGmeligMeyling:
Fixed commitid expression in delivery.ftl


LiamClark:
LGTM

PR #283, opened 2016-04-16
Batch releases, fixes #99

### LINKED ISSUE #99: Hide/release grades for an assignment
A way to hide the grades for the students during the grading process and then release all the grades together.

#81

Consists of three parts:
- [x] Remove code for sending email after grade
- [x] Add boolean flag to assignment, create migration for adding this field, default for existing assignments to `true`
- [x] Hide delivery results in all places of Devhub if this flag is `false`

LiamClark:
Grades were always released as soon as the ta filled them in, we want to manage this and release the grades at the same time, this pr fixes that.

We no longer send an email when the grade is uploaded and the email field is completely gone now from the ProjectAssignmentResource.

The ProjectAssignmentResource checks wether the user is a TA or admin and always shows them the grade.

If the user is a student he will only see it once the assignment grades have been released. 

Potential problem it adds a flag which is checked in the freemarker template, this flag get's used in the delivery render macro, if it get's called anywhere else this will break the other call site. 

@JWGmeligMeyling do you know about such another callsite or a nicer way to do this?

finally the edit assignment form shows a release grade check mark if the assignment already exists, because making an assignment that immediately releases the grades makes little sense.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AssignmentsResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AssignmentsResource.java
@@ -312,7 +312,8 @@ public class AssignmentsResource extends Resource {
                                    @PathParam("assignmentId") long assignmentId,
                                    @FormParam("name") String name,
                                    @FormParam("summary") String summary,
-                                   @FormParam("due-date") String dueDate) {
+                                   @FormParam("due-date") String dueDate,
+                                   @FormParam("release") Boolean release) {
 
         CourseEdition course = courses.find(courseCode, editionCode);
 
@@ -323,7 +324,7 @@ public class AssignmentsResource extends Resource {
         Assignment assignment = assignmentsDAO.find(course, assignmentId);
         assignment.setName(name);
         assignment.setSummary(summary);
-
+        assignment.setGradesReleased(release);
 
         if(!Strings.isNullOrEmpty(dueDate)) {
             SimpleDateFormat simpleDateFormat = new SimpleDateFormat(DATE_FORMAT);
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AssignmentsResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AssignmentsResource.java
@@ -313,7 +313,7 @@ public class AssignmentsResource extends Resource {
                                    @FormParam("name") String name,
                                    @FormParam("summary") String summary,
                                    @FormParam("due-date") String dueDate,
-                                   @FormParam("release") Boolean release) {
+                                   @FormParam("release") String release) {
 
         CourseEdition course = courses.find(courseCode, editionCode);
 
@@ -325,9 +325,9 @@ public class AssignmentsResource extends Resource {
         assignment.setName(name);
         assignment.setSummary(summary);
 
-        if(release != null) {
-            assignment.setGradesReleased(release);
-        }
+        //a little bit ugly, form params don't work nicely with checkboxes.
+        boolean releaseStatus = release != null && release.equals("on");
+        assignment.setGradesReleased(releaseStatus);
 
         if(!Strings.isNullOrEmpty(dueDate)) {
             SimpleDateFormat simpleDateFormat = new SimpleDateFormat(DATE_FORMAT);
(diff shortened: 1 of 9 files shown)


src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AssignmentsResource.java
```diff
@@ -324,7 +324,10 @@ public Response editAssignment(@Context HttpServletRequest request,
         Assignment assignment = assignmentsDAO.find(course, assignmentId);
         assignment.setName(name);
         assignment.setSummary(summary);
-        assignment.setGradesReleased(release);
+
+        if(release != null) {
+            assignment.setGradesReleased(release);
```
JWGmeligMeyling: Does this break the ability to "unpublish" the grades?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AssignmentsResource.java
```diff
@@ -324,6 +325,9 @@ public Response editAssignment(@Context HttpServletRequest request,
         assignment.setName(name);
         assignment.setSummary(summary);
 
+        if(release != null) {
```
JWGmeligMeyling: Does this break the ability to "unpublish" the grades?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AssignmentsResource.java
```diff
@@ -312,7 +312,8 @@ public Response editAssignment(@Context HttpServletRequest request,
                                    @PathParam("assignmentId") long assignmentId,
                                    @FormParam("name") String name,
                                    @FormParam("summary") String summary,
-                                   @FormParam("due-date") String dueDate) {
+                                   @FormParam("due-date") String dueDate,
+                                   @FormParam("release") Boolean release) {
```
JWGmeligMeyling: Maybe `@DefaultValue("false")` is a better idea here. How does the checkbox serialise in a plain html form? Is this serialisable as boolean? ( is the outputted value `"true"` or something else?)

src/main/resources/templates/components/delivery.ftl
```diff
@@ -69,9 +69,8 @@
     </ul>
     [/#if]
 
-
     [#assign review = delivery.getReview()!]
-    [#if review?? && review?has_content && (review.getGrade()?? || review.getCommentary()??)]
+    [#if review?? && seeGrade && review?has_content && (review.getGrade()?? || review.getCommentary()??)]
```
JWGmeligMeyling: Yes, not ideal, but we will change that when we start transitioning to a more MVC alike pattern, and not in this PR 😄

src/test/java/nl/tudelft/ewi/devhub/server/web/resources/ProjectAssignmentsResourceTest.java
```diff
@@ -0,0 +1,65 @@
...
+    @Mock
+    private Group group;
+
+    @Mock
+    private User user;
+
+    @Mock
+    private Assignment assignment;
+
+    private ProjectAssignmentsResource resource;
```
JWGmeligMeyling: `@InjectMocks` can be used here (not necessary to change this)

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/AssignmentsResource.java
```diff
@@ -325,9 +325,9 @@ public Response editAssignment(@Context HttpServletRequest request,
         assignment.setName(name);
         assignment.setSummary(summary);
 
-        if(release != null) {
-            assignment.setGradesReleased(release);
-        }
+        //a little bit ugly, form params don't work nicely with checkboxes.
+        boolean releaseStatus = release != null && release.equals("on");
```
LiamClark: This fixes the unpublishing issue, I misunderstood the FormParam at first this is ugly but works, they don't do check boxes that nicely.

JWGmeligMeyling:
Will be included in the 3.2.2 milestone (hopefully the weekend, but anyway before the first assignment)

JWGmeligMeyling:
I'll make this up to date with master tomorrow.

PR #306, opened 2016-04-21
Initial version of testing the project backend, fixes #295

### LINKED ISSUE #295: Provisioning repository some times fail, retries result in group number gaps

JWGmeligMeyling:
Work with @LiamClark to reproduce the problem of issue #295 . This PR only proved that what we thought that was going wrong, is actually working fine. Through the debugging work on the issue, we managed to address parts of the issue in the git-server (https://github.com/devhub-tud/git-server/pull/73). This however does still not explain why the provisioning failed, and the repository had to be removed in the first place.

It can be the case that provisioning failed because a previous failure was not cleaned up, and the successive group creation failed because the repository was already existing. This however does not explain why both groups received the same group number. The group number should always increment, as we can also see from the current group number gaps.


JWGmeligMeyling:
Closed as we assume this bug is to be fixed and the test only shows that the existing behaviour should be correct...

PR #336, opened 2016-05-05
Implement a proper test ViewState for assignments

DouweKoopmans:
Implement a view for Assignments with proper asserts and add a basic test for the assignment page.

### DIFF
--- a/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentReviewTest.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentReviewTest.java
@@ -87,7 +87,7 @@ public class AssignmentReviewTest extends WebTest{
 
         assertEquals(modelDelivery.getCreatedUser().getName(), viewAssignment.getAuthor());
         assertEquals(modelAssignment.getName(), viewAssignment.getName());
-//        assertEquals(modelDelivery.getReview().getState(), viewAssignment.getStatus());
+        assertEquals(modelDelivery.getReview().getState(), viewAssignment.getStatus());
         assertEquals(modelDelivery.getReview().getReviewUser().getName(), viewAssignment.getReview().getReviewer());
     }
 
--- a/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentReviewTest.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentReviewTest.java
@@ -1,0 +1,106 @@
+package nl.tudelft.ewi.devhub.webtests;
+
+import com.google.inject.Inject;
+import nl.tudelft.ewi.devhub.server.database.controllers.CourseEditions;
+import nl.tudelft.ewi.devhub.server.database.controllers.Deliveries;
+import nl.tudelft.ewi.devhub.server.database.controllers.Groups;
+import nl.tudelft.ewi.devhub.server.database.entities.Assignment;
+import nl.tudelft.ewi.devhub.server.database.entities.CourseEdition;
+import nl.tudelft.ewi.devhub.server.database.entities.Delivery;
+import nl.tudelft.ewi.devhub.server.database.entities.Group;
+import nl.tudelft.ewi.devhub.webtests.utils.WebTest;
+import nl.tudelft.ewi.devhub.webtests.views.DeliveryReviewView;
+import org.junit.Test;
+
+import java.text.ParseException;
+
+import javax.persistence.EntityManager;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by Douwe Koopmans on 9-5-16.
+ */
+public class AssignmentReviewTest extends WebTest{
+    private static final String ASSISTANT_USERNAME = "assistant1";
+    private static final String ASSISTANT_PASSWORD = "assistant1";
+
+    @Inject
+    private CourseEditions courseEditions;
+
+    @Inject
+    private Deliveries deliveries;
+
+    @Inject
+    private EntityManager entityManager;
+
+    @Inject
+    private Groups groups;
+
+    @Test
+    public void testDeliverable() throws ParseException {
+        final DeliveryReviewView view = getDeliveryReviewView();
+
+        final CourseEdition course = courseEditions.find(1);
+        final Group group = groups.find(course).get(0);
+        final Delivery modelDelivery = deliveries.find(group,2L);
+        final Assignment modelAssignment = modelDelivery.getAssignment();
+        final DeliveryReviewView.Assignment viewAssignment = view.getAssignment();
+
+        assertEquals(modelDelivery.getCreatedUser().getName(), viewAssignment.getAuthor());
+        assertEquals(modelAssignment.getName(), viewAssignment.getName());
+        assertEquals(modelDelivery.getState(), viewAssignment.getStatus());
+    }
+
+    @Test
+    public void testReviewForm() throws ParseException {
+        final DeliveryReviewView view = getDeliveryReviewView();
+
+        final DeliveryReviewView.DeliveryForm deliveryForm = view.getDelivery();
+        deliveryForm.setGrade("4.0");
+        deliveryForm.setState("Rejected");
+        deliveryForm.setCommentary("Hello World!");
+
+        assertTrue(deliveryForm.getGrade() == 4.0D);
+        assertEquals(Delivery.State.REJECTED, deliveryForm.getSelectedState());
+        assertEquals("Hello World!", deliveryForm.getCommentary());
+    }
+
+    @Test
+    public void testSubmitReview() throws ParseException{
+        final DeliveryReviewView view = getDeliveryReviewView();
+
+        final DeliveryReviewView.DeliveryForm deliveryForm = view.getDelivery();
+        deliveryForm.setCommentary("Hello World!");
+        deliveryForm.setGrade("7.0");
+        deliveryForm.setState("Approved");
+        deliveryForm.click();
+
+        entityManager.clear();
+
+        final CourseEdition course = courseEditions.find(1);
+        final Group group = groups.find(course).get(0);
+        final Delivery modelDelivery = deliveries.find(group,2L);
+        final Assignment modelAssignment = modelDelivery.getAssignment();
+        final DeliveryReviewView.Assignment viewAssignment = view.getAssignment();
+
+        assertEquals(modelDelivery.getCreatedUser().getName(), viewAssignment.getAuthor());
+        assertEquals(modelAssignment.getName(), viewAssignment.getName());
+//        assertEquals(modelDelivery.getReview().getState(), viewAssignment.getStatus());
+        assertEquals(modelDelivery.getReview().getReviewUser().getName(), viewAssignment.getReview().getReviewer());
+    }
+
+    private DeliveryReviewView getDeliveryReviewView() throws ParseException {
+        return openLoginScreen()
+                .login(ASSISTANT_USERNAME, ASSISTANT_PASSWORD)
+                .listAssistingCourses()
+                .get(0).click()
+                .listGroups()
+                .get(0).click()
+                .toAssignmentView()
+                .listAssignments()
+                .get(1).click()
+                .getAssignment().click();
+    }
+}
--- a/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
@@ -12,6 +12,9 @@ import static org.junit.Assert.assertEquals;
 
 public class AssignmentTest extends WebTest {
 
+	private static final String ASSISTANT_USERNAME = "assistant1";
+	private static final String ASSISTANT_PASSWORD = "assistant1";
+
 	@Test
 	public void testListAssignments() {
 		AssignmentsView view = openLoginScreen()
@@ -32,11 +35,8 @@ public class AssignmentTest extends WebTest {
 
 	@Test
 	public void testAssignment() throws ParseException {
-		final String net_id = "assistant1";
-		final String password = "assistant1";
-
 		final AssignmentView view = openLoginScreen()
-				.login(net_id, password)
+				.login(ASSISTANT_USERNAME, ASSISTANT_PASSWORD)
 				.listAssistingCourses()
 				.get(0).click()
 				.listGroups()
@@ -47,7 +47,8 @@ public class AssignmentTest extends WebTest {
 
 		final AssignmentView.Assignment assignment = view.getAssignment();
         assertEquals("Student Two", assignment.getAuthor());
-        // TODO: 4-5-16 assert date is equal to current date
+		assertEquals("Product vision", assignment.getName());
+        assertEquals("Submitted", assignment.getStatus());
     }
 	
 }
--- a/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
@@ -60,7 +60,7 @@ public class AssignmentTest extends WebTest {
 				.get(0).click()
 				.toAssignmentView()
 				.listAssignments()
-				.get(0).click();
+				.get(1).click();
 
         final CourseEdition course = courseEditions.find(1);
         final Group group = groups.find(course).get(0);
--- a/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
@@ -1,5 +1,13 @@
 package nl.tudelft.ewi.devhub.webtests;
 
+import com.google.inject.Inject;
+import nl.tudelft.ewi.devhub.server.database.controllers.CourseEditions;
+import nl.tudelft.ewi.devhub.server.database.controllers.Deliveries;
+import nl.tudelft.ewi.devhub.server.database.controllers.Groups;
+import nl.tudelft.ewi.devhub.server.database.entities.Assignment;
+import nl.tudelft.ewi.devhub.server.database.entities.CourseEdition;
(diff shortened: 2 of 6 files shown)


src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
```diff
@@ -28,5 +29,25 @@ public void testListAssignments() {
 		assertEquals(assignment.getName(), "Product vision");
 		assertEquals(assignment.getNumber(), "1");
 	}
+
+	@Test
+	public void testAssignment() throws ParseException {
+		final String net_id = "assistant1";
```
JWGmeligMeyling: Please store these values in a static variable, such as `ASSISTANT_USERNAME` :smile:

src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
```diff
@@ -28,5 +29,25 @@ public void testListAssignments() {
...
+				.listAssistingCourses()
+				.get(0).click()
+				.listGroups()
+				.get(0).click()
+				.toAssignmentView()
+				.listAssignments()
+				.get(0).click();
+
+		final AssignmentView.Assignment assignment = view.getAssignment();
+        assertEquals("Student Two", assignment.getAuthor());
```
JWGmeligMeyling: We should add some more assertions here, are we looking at the expected assignment?

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,59 @@
...
+        assertTrue(currentPathStartsWith("/courses/ti1705/TI1705/groups/"));
+    }
+
+    public Assignment getAssignment() throws ParseException {
+        String author;
+        Date date;
+        WebElement submit = getDriver().findElement(By.cssSelector("a[href^=\"deliveries\"]"));
+
+        final List<WebElement> elements = getDriver().findElements(By.tagName("dd"));
+        author = elements.get(0).getText();
```
JWGmeligMeyling: Please also initialize the variable here as well. Otherwise it is a non-constant variable.

JWGmeligMeyling:
Nice, 0.7% more coverage. Way to go :smile: Add some more assertions, so we know for sure that the test doesn't pass for just  any view. Add some get methods to the view which allow you do do assertions against the actual `Assignment` and `Delivery` objects.

src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentTest.java
```diff
@@ -1,16 +1,20 @@
...
+import java.text.ParseException;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
 public class AssignmentTest extends WebTest {
 
+	private static final String ASSISTANT_USERNAME = "assistant1";
+	private static final String ASSISTANT_PASSWORD = "assistant1";
+
```
JWGmeligMeyling: You can add `@Inject CourseEditions courseEditions` and `@Inject Assignments assignments` here and then do 
`courseEditions.find(1).get()` and `assignments.find(courseEdition, 1)` to retrieve the actual `CourseEdition` and `Assignment` instances, so you would not have to hard code strings as `Student Two`, `Product Vision` and `Submitted` (just as a mere example on how you can get more control over the various parts in Devhub in your test)

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,59 @@
...
+        super.invariant();
+        assertNotNull(getDriver().findElement(RECENT_ASSIGNMENTS_HEADER));
+        assertTrue(currentPathStartsWith("/courses/ti1705/TI1705/groups/"));
+    }
+
+    public Assignment getAssignment() throws ParseException {
+        WebElement submit = getDriver().findElement(By.cssSelector("a[href^=\"deliveries\"]"));
+
+        final List<WebElement> elements = getDriver().findElements(By.tagName("dd"));
+        final String author = elements.get(0).getText();
```
JWGmeligMeyling: Please move this to a method `getAuthor()` in your `Assignment` class.

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,59 @@
...
+        final Date date = new SimpleDateFormat("EEEE dd MMMM yyyy k:m", Locale.US).parse(elements.get(1).getText());
```
JWGmeligMeyling: Please move this to a method `getDueDate()` in your Assignment class.

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,59 @@
...
+        final String name = getDriver().findElement(By.cssSelector("div .col-md-offset-2.col-md-10 h4")).getText();
```
JWGmeligMeyling: Please move this to a method `getName()` in your Assignment class.

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,59 @@
...
+        final String status = getDriver().findElement(By.cssSelector("span.label")).getText();
```
JWGmeligMeyling: Please move this to a method `getStatus()` in your Assignment class. Try to have it return `Delivery.State` (perhaps using `Delivery.State.valueOf([...].getText().toUpperCase())`

JWGmeligMeyling:
I have just added some more comments. Besides these, could you add another test case for when you click the next anchor?

JWGmeligMeyling:
:+1: This is exactly what I meant :smile:

src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentReviewTest.java
```diff
@@ -0,0 +1,106 @@
...
+        assertEquals(modelDelivery.getState(), viewAssignment.getStatus());
+    }
+
+    @Test
+    public void testReviewForm() throws ParseException {
+        final DeliveryReviewView view = getDeliveryReviewView();
+
+        final DeliveryReviewView.DeliveryForm deliveryForm = view.getDelivery();
+        deliveryForm.setGrade("4.0");
+        deliveryForm.setState("Rejected");
```
JWGmeligMeyling: Could have been `WordUtils.capitalize(Delivery.State.REJECTED)` so the test is more stable (will cause compilation errors) if we at some point change the Delivery API.

src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentReviewTest.java
```diff
@@ -0,0 +1,106 @@
...
+        deliveryForm.setCommentary("Hello World!");
```
JWGmeligMeyling: We should store this in a static variable so it can be reused in the test (and if we change it, we don't have to change statements elsewhere in the test)

src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentReviewTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+        final CourseEdition course = courseEditions.find(1);
+        final Group group = groups.find(course).get(0);
+        final Delivery modelDelivery = deliveries.find(group,2L);
+        final Assignment modelAssignment = modelDelivery.getAssignment();
+        final DeliveryReviewView.Assignment viewAssignment = view.getAssignment();
+
+        assertEquals(modelDelivery.getCreatedUser().getName(), viewAssignment.getAuthor());
+        assertEquals(modelAssignment.getName(), viewAssignment.getName());
+//        assertEquals(modelDelivery.getReview().getState(), viewAssignment.getStatus());
```
JWGmeligMeyling: Remove this commented line? Or should we fix it?

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,126 @@
...
+                return null;
+            }
+
+            return new SimpleDateFormat("EEEE dd MMMM yyyy k:m", Locale.US).parse(footer.split(" on ")[1].trim());
+        }
+
+        @Nullable
+        private String getFooter() {
+            try {
+                return getDriver().findElement(By.cssSelector("footer.small")).getText();
```
JWGmeligMeyling: Catching exceptions in tests is actually quite bad! Two things can happen:
* The element was expected to be not existing, a false warning is being logged.
* The element was not present while it was expected, but the test still passes
Which behaviour is intended is very hard to read and more importantly, error prone.

Try:
```java
getDriver().findElements(By.id("btn-close")).stream().findAny().isPresent();
```

src/test/java/nl/tudelft/ewi/devhub/webtests/views/DeliveryReviewView.java
```diff
@@ -0,0 +1,202 @@
...
+                    getDriver().findElements(By.tagName("dd")).get(1).getText());
+        }
+    }
+
+    @Data
+    public class Review {
+        public String getReviewer() {
+            final String footer = getFooter();
+            if (footer == null) {
+                return null;
```
JWGmeligMeyling: Can this ever happen?

src/test/java/nl/tudelft/ewi/devhub/webtests/views/DeliveryReviewView.java
```diff
@@ -0,0 +1,202 @@
...
+            try {
+                return getDriver().findElement(By.cssSelector("footer.small")).getText();
+            } catch (Exception e) {
+                return null;
+            }
+        }
+    }
+
+
+    @Data
```
JWGmeligMeyling: You can also remove this annotation if you don't want the getters

JWGmeligMeyling:
PR looks good! A few small remarks!

src/test/java/nl/tudelft/ewi/devhub/webtests/AssignmentReviewTest.java
```diff
@@ -0,0 +1,106 @@
...
+    @Test
+    public void testReviewForm() throws ParseException {
+        final DeliveryReviewView view = getDeliveryReviewView();
+
+        final DeliveryReviewView.DeliveryForm deliveryForm = view.getDelivery();
+        deliveryForm.setGrade("4.0");
+        deliveryForm.setState("Rejected");
+        deliveryForm.setCommentary("Hello World!");
+
+        assertTrue(deliveryForm.getGrade() == 4.0D);
```
JWGmeligMeyling: I missed this in the first review round, but you should not use `==` for double equality. This is because doubles are not exact values and cannot be compared without fault tolerance. This statement is likely to fail in a lot of correct cases. The correct assertion is `assertEquals(4.0, deliveryForm.getGrade(), 1e-4`) where `1e-4` is the precision (can be smaller as well)

src/test/java/nl/tudelft/ewi/devhub/webtests/views/DeliveryReviewView.java
```diff
@@ -0,0 +1,196 @@
...
+
+
+    public DeliveryReviewView(WebDriver driver) {
+        super(driver);
+    }
+
+    @Override
+    public void invariant() {
+        super.invariant();
+        Assert.assertTrue(currentPathStartsWith("/courses/ti1705/TI1705/groups/"));
```
JWGmeligMeyling: Will be better if we make the tests less course specific, so perhaps a `currentPathContains('assignment')

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,126 @@
...
+        public Delivery.State getStatus() {
+            return Delivery.State.valueOf(getDriver().findElement(By.cssSelector("span.label")).getText().toUpperCase());
+        }
+
+        public Date getDueDate() throws ParseException {
+            return new SimpleDateFormat("EEEE dd MMMM yyyy k:m", Locale.US).parse(
+                    getDriver().findElements(By.tagName("dd")).get(1).getText());
+        }
+
+        public DeliveryReviewView click() {
```
JWGmeligMeyling: Its good to create an `Assignment` class that is able to resolve a few properties of the assignment. However, navigating to the `DeliveryReviewView` is specific for the assignment submission/delivery, and not for the assignment. Furthermore, the method name `click` is not very descriptive for going to the review view here (in contract to `getAssignments().get().click()` where you at least have some idea that you will navigate to the assignments page).

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Data
+    public class Review {
+        public String getReviewer() {
+            final String footer = getFooter();
+            if (footer == null) {
+                return null;
+            }
+
+            return footer.split(" on ")[0].trim();
```
JWGmeligMeyling: Could benefit a lot from some `Optional` passing 😄

src/test/java/nl/tudelft/ewi/devhub/webtests/views/AssignmentView.java
```diff
@@ -0,0 +1,126 @@
...
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Created by Douwe Koopmans on 4-5-16.
+ */
+public class AssignmentView extends ProjectSidebarView {
+
+    private static final By RECENT_ASSIGNMENTS_HEADER = By.xpath("//h4[starts-with(normalize-space(.), 'Assignments')]");
```
JWGmeligMeyling: This header is not present, so the invariant fails. (`Assignment#getName()` is the correct value here). This invalid assertion should be removed.

src/test/java/nl/tudelft/ewi/devhub/webtests/views/DeliveryReviewView.java
```diff
@@ -0,0 +1,196 @@
...
+            invariant();
+            final WebElement element = getDriver().findElement(STATE_SELECTOR);
+            final Select select = new Select(element);
+
+            final Optional<WebElement> optionalNewStateElement = select.getOptions()
+                    .stream()
+                    .filter(webElement -> webElement.getText().equals(newState)).findAny();
+
+            if (optionalNewStateElement.isPresent()) {
+                optionalNewStateElement.get().click();
```
JWGmeligMeyling: The `get` will throw an `NoSuchElementException` if not present. As you expect the element to exist, and want to throw an exception when it doesn't anyway, you could just write:

```java
select.getOptions()
  .stream()
  .filter(webElement -> webElement.getText().equals(newState)).findAny()
  .get().click();
```

src/test/java/nl/tudelft/ewi/devhub/webtests/views/DeliveryReviewView.java
```diff
@@ -0,0 +1,196 @@
...
+
+        public double getGrade() {
+            return Double.parseDouble(getDriver().findElement(GRADE_SELECTOR).getAttribute("value"));
+        }
+
+        public void setGrade(String newGrade) throws IllegalArgumentException {
+            invariant();
+
+            if (Double.parseDouble(newGrade) < 1D || Double.parseDouble(newGrade) > 10D) {
+                throw new IllegalArgumentException(newGrade + " is not a valid grade, grades must be between 1.0 and 10");
```
JWGmeligMeyling: This is actually not true. We have designed the grading system such that it supports international grading systems, and does not force you to be within `1` and `10`. Therefore we have the separation of `grade` and `state` (any of `rejected`, `approved`, `disapproved`). How the `grade` maps to the `state` (or the computed number of points through the rubrics) is course specific.

src/test/java/nl/tudelft/ewi/devhub/webtests/views/DeliveryReviewView.java
```diff
@@ -0,0 +1,196 @@
...
+         * options:
+         * - Submitted
+         * - Rejected
+         * - Approved
+         * - Disapproved
+         *
+         * @param newState
+         * @throws IllegalArgumentException
+         */
+        public void setState(String newState) throws IllegalArgumentException {
```
JWGmeligMeyling: By accepting a `Delivery.State` enum value, you eliminate the need of a descriptive list of the various applicable options :wink:

JWGmeligMeyling:
@DouweKoopmans I was so kind to fix my outstanding feedback as I want to move on to Part 2 now 😄  Please do read my comments and my changes, and then hit merge 😉

JWGmeligMeyling:
Coverage is running again, and your branch is up to date with master, please review my remarks (that I already fixed for you) and merge :smile:

src/test/java/nl/tudelft/ewi/devhub/webtests/views/DeliveryReviewView.java
DouweKoopmans: this method doesn't exist yet, shall i add it to the View class?

JWGmeligMeyling: It is added since I merged master into your branch I believe :wink:

src/test/java/nl/tudelft/ewi/devhub/webtests/views/DeliveryReviewView.java
```diff
@@ -134,7 +130,8 @@ public State getSelectedState() {
             return State.valueOf(select.getFirstSelectedOption().getText().toUpperCase());
         }
 
-        public void setState(Delivery.State state) throws IllegalArgumentException {
+        @SuppressWarnings("OptionalGetWithoutIsPresent")
```
JWGmeligMeyling: Sometimes warnings just don't make any sense.. :wink:

PR #339, opened 2016-05-09
Pullrequest template fixes

AZWN:
In this branch I fixed 3 issues regarding the pull request frontend:

1. In The message after a merge succeeded it said that the branch was not merged, while in fact it was.
2. In The message after a merge failed a single quote was not parsed correctly by the translator.
3. When going back after creating a pull-request, there was still a button "Create pull request", which gave a 404 error when clicked. This is fixed by adding no-store to the Cache-Control header.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
@@ -61,7 +61,9 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.CacheControl;
 import javax.ws.rs.core.Context;
+import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


src/main/resources/i18n/devhub_en.properties
```diff
@@ -244,9 +244,10 @@ pull-request.open.message = Hey! You can merge the pull request by clicking the
 pull-request.merging = Merging...
 pull-request.merged = Merged
 pull-request.failed-to-merge = Failed to merge
-pull-request.failed-to-merge.message = We can’t automatically merge this pull request due to merge conflicts. \
+pull-request.failed-to-merge.message = We can\\''t automatically merge this pull request due to merge conflicts. \
```
JWGmeligMeyling: Shouldn't this be `\\'`?

AZWN: Whoops, that is old. The \\ can completely be removed.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
```diff
@@ -159,6 +160,7 @@ public Response showProjectOverview(@Context HttpServletRequest request,
 	}
 
 	@GET
+	@Cache(noStore = true)
```
JWGmeligMeyling: We like it non-verbose :smile:

PR #354, opened 2016-05-12
endpoint for autograding

LiamClark:
simple end point for the last delivery of a group for auto grading purposes, intelij screwed up the imports. not enough time to fix it right now.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/ProjectAssignmentsResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/ProjectAssignmentsResource.java
@@ -1,5 +1,14 @@
 package nl.tudelft.ewi.devhub.server.web.resources;
 
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
+import com.google.inject.persist.Transactional;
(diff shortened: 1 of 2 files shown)


src/main/java/nl/tudelft/ewi/devhub/server/web/resources/ProjectAssignmentsResource.java
```diff
@@ -414,7 +400,32 @@ public void persistMasteries(
...
+    public static class AutoGradeResult {
+        private List<Long> ids;
+    }
+
+
+    @POST
+    @Transactional
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("{assignmentId}/last-delivery/auto-masteries")
+    public void uploadAutoGradingGrade(@PathParam("assignmentId") long assignmentId, AutoGradeResult result) {
```
JWGmeligMeyling: I'll merge this anyway as this is just for a specific purpose and we will change it anyway, but I don't really like an array of of ints. It makes it a little bit difficult to extend if you add new fields (or start using for example compound identifiers). So it makes your API less open for extension. Also its not completely according to the REST principles (where you usually post the entities you want to connect). By using the `Mastery` object, the request did not get more complex, yet more sensible (`[{id : 5}]` over `[5]`), and you could have eliminated 6 lines of boilerplate that are now required to transform the list 😄

JWGmeligMeyling: *Plus you would have the option to make the client side type safe if you shared the data object/interface*

PR #359, opened 2016-05-14
Flatten directory tree, fixes #347

### LINKED ISSUE #347: Folders with 1 subfolder should be collapsed in the file view
When a folder has only 1 subfolder (no files) it can be truncated. E.g. `src/test/java` has 1 subfolder `nl`. Now you have to continue clicking through it, while the GitHub behavior is to directly go from `src/test` to `src/test/java/nl/...`. This saves a lot of clicks :)

JWGmeligMeyling:
screen shot 2016-05-14 at 14 53 15

Fixes #347


JWGmeligMeyling:
The `FlattenFolderTree` should be tested against a mocked (Mockito) `CommitApi `.

PR #374, opened 2016-05-15
Boilerplate for equals hashcode tests

JWGmeligMeyling:


JWGmeligMeyling:
`junit-addons` adds some transitive dependencies which conflict with our dependencies. As the dependencies for `junit-addons`, we should exclude them from the `pom.xml`. Do not merge this just yet :wink:

PR #384, opened 2016-05-19
IssueTracker, fixes #148

### LINKED ISSUE #148: Issue tracker
See also https://github.com/devhub-tud/devhub/blob/master/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/AbstractIssue.java#L34 and how Pull Requests are currently handled.

AZWN:
Implementing a basic issue tracker. Still in progress.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Issues.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Issues.java
@@ -48,4 +48,11 @@ public class Issues extends Controller<Issue> {
 			.list(issue);
 	}
 
+	@Transactional
+	public List<Issue> findIssueById(final RepositoryEntity repo, final long id){
+		return query().from(issue)
+			.where(issue.repository.eq(repo).and(issue.issueId.eq(id)))
+			.list(issue);
+	}
+
 }
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Issues.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Issues.java
@@ -21,7 +21,7 @@ public class Issues extends Controller<Issue> {
 	}
 	
 	@Transactional
-	public List<Issue> findIssues(final RepositoryEntity repo, User user){
+	public List<Issue> findAssignedIssues(final RepositoryEntity repo, User user){
 		return query().from(issue)
 			.where(issue.repository.eq(repo).and(issue.assignee.eq(user)))
 			.list(issue);
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/AbstractIssue.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/AbstractIssue.java
@@ -77,7 +77,7 @@ public abstract class AbstractIssue implements Event, Base {
 	@Temporal(TemporalType.TIMESTAMP)
 	private Date closed;
 
-	// Left away @NotEmpty for now, since existing tests should be updated for it to work
+	@NotEmpty
 	@Column(name="title")
 	private String title;
 
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
@@ -8,6 +8,7 @@ import javax.persistence.Table;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
+import lombok.ToString;
 
 
 @Data
@@ -15,6 +16,7 @@ import lombok.NoArgsConstructor;
 @Table(name="repository_issues")
 @NoArgsConstructor
 @EqualsAndHashCode(callSuper = true)
+@ToString(callSuper=true, includeFieldNames=true)
 public class Issue extends AbstractIssue {
 
 	@Override
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
@@ -1,14 +1,20 @@
 package nl.tudelft.ewi.devhub.server.database.entities.issues;
 
 import java.net.URI;
+import java.util.List;
 
+import javax.persistence.CascadeType;
 import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.OneToMany;
+import javax.persistence.OrderBy;
 import javax.persistence.Table;
 
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 import lombok.ToString;
+import nl.tudelft.ewi.devhub.server.database.entities.comments.IssueComment;
 
 
 @Data
@@ -23,5 +29,9 @@ public class Issue extends AbstractIssue {
 	public URI getURI() {
 		return getRepository().getURI().resolve("issue/" + getIssueId() + "/");
 	}
+
+	@OrderBy("timestamp ASC")
+	@OneToMany(mappedBy = "issues", fetch = FetchType.LAZY, cascade = {CascadeType.DETACH, CascadeType.REMOVE}, orphanRemoval = true)
+	private List<IssueComment> comments;
 	
 }
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
@@ -5,12 +5,16 @@ import java.net.URI;
 import javax.persistence.Entity;
 import javax.persistence.Table;
 
+import lombok.Data;
+import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 
 
+@Data
 @Entity
 @Table(name="repository_issues")
 @NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
 public class Issue extends AbstractIssue {
 
 	@Override
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
@@ -31,7 +31,7 @@ public class Issue extends AbstractIssue {
 	}
 
 	@OrderBy("timestamp ASC")
-	@OneToMany(mappedBy = "issues", fetch = FetchType.LAZY, cascade = {CascadeType.DETACH, CascadeType.REMOVE}, orphanRemoval = true)
+	@OneToMany(mappedBy = "issue", fetch = FetchType.LAZY, cascade = {CascadeType.DETACH, CascadeType.REMOVE}, orphanRemoval = true)
 	private List<IssueComment> comments;
 	
 }
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
@@ -5,15 +5,11 @@ import java.net.URI;
 import javax.persistence.Entity;
 import javax.persistence.Table;
 
-import lombok.Data;
-import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 
 
-@Data
 @Entity
 @Table(name="repository_issues")
-@EqualsAndHashCode(callSuper = true)
 @NoArgsConstructor
 public class Issue extends AbstractIssue {
 
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
@@ -79,8 +79,8 @@ public abstract class AbstractProjectIssueResource extends AbstractIssueResource
 		Map<String, Object> parameters = getBaseParameters();
 		
 		parameters.put("repository", repository);
-		parameters.put("openPullRequests", openIssues);
-		parameters.put("closedPullRequests", closedIssues);
+		parameters.put("openIssues", openIssues);
+		parameters.put("closedIssues", closedIssues);
 
 		List<Locale> locales = Collections.list(request.getLocales());
 		return display(templateEngine.process("courses/assignments/group-issues.ftl", locales, parameters));
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
@@ -15,7 +15,6 @@ import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
-import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -27,11 +26,10 @@ import nl.tudelft.ewi.devhub.server.backend.mail.CommentMailer;
 import nl.tudelft.ewi.devhub.server.database.controllers.IssueComments;
 import nl.tudelft.ewi.devhub.server.database.controllers.Issues;
 import nl.tudelft.ewi.devhub.server.database.controllers.Users;
-import nl.tudelft.ewi.devhub.server.database.entities.Group;
 import nl.tudelft.ewi.devhub.server.database.entities.RepositoryEntity;
 import nl.tudelft.ewi.devhub.server.database.entities.User;
+import nl.tudelft.ewi.devhub.server.database.entities.comments.IssueComment;
 import nl.tudelft.ewi.devhub.server.database.entities.issues.Issue;
-import nl.tudelft.ewi.devhub.server.database.entities.issues.PullRequest;
 import nl.tudelft.ewi.devhub.server.web.errors.UnauthorizedException;
 import nl.tudelft.ewi.devhub.server.web.resources.Resource;
 import nl.tudelft.ewi.devhub.server.web.templating.TemplateEngine;
@@ -200,6 +198,24 @@ public abstract class AbstractProjectIssueResource extends AbstractIssueResource
 		}
 		
 		issues.merge(issue);
+
+		return redirect(issue.getURI());
+	}
+	
+	@POST
+	@Transactional
+	@Path("/issue/{issueId}/comment")
+	public Response addComment(@Context HttpServletRequest request, 
+			@PathParam("issueId") long issueId,
+			@FormParam("content") String content) throws IOException {
+		
+		Issue issue = issues.findIssueById(getRepositoryEntity(), issueId).get(0);
+		IssueComment comment = new IssueComment();
+		comment.setContent(content);
+		comment.setIssue(issue);
+		comment.setUser(currentUser);
+		
+		issueComments.persist(comment);
 
 		return redirect(issue.getURI());
 	}
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
@@ -24,6 +24,7 @@ import lombok.extern.slf4j.Slf4j;
 import nl.tudelft.ewi.devhub.server.backend.CommentBackend;
 import nl.tudelft.ewi.devhub.server.backend.IssueBackend;
 import nl.tudelft.ewi.devhub.server.backend.mail.CommentMailer;
+import nl.tudelft.ewi.devhub.server.database.controllers.IssueComments;
 import nl.tudelft.ewi.devhub.server.database.controllers.Issues;
 import nl.tudelft.ewi.devhub.server.database.controllers.Users;
 import nl.tudelft.ewi.devhub.server.database.entities.Group;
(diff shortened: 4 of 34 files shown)


src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
```diff
@@ -0,0 +1,42 @@
...
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import nl.tudelft.ewi.devhub.server.database.entities.User;
+
+
+@Data
+@Entity
+@Table(name="issues")
```
JWGmeligMeyling: Maybe call this `repository_issues` while we still can.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.validation.constraints.NotNull;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import nl.tudelft.ewi.devhub.server.database.entities.User;
+
+
+@Data
+@Entity
```
JWGmeligMeyling: Please read a little bit on [JPA Inheritance](https://en.wikibooks.org/wiki/Java_Persistence/Inheritance) if you're not yet familiar with that. I am thinking whether it would be a good idea to use `InheritanceType.SINGLE_TABLE` or `InheritanceType.SINGLE_TABLE`, instead of making `AbstractIssue` a `@MappedSuperclass`. The benefit of this is that we can make foreign keys relation ships to the `AbstractIssue`, and by that means make the `PullRequestComment` an `AbstractIssueComment`, without having to introduce a separate `IssueComment` (plus a data access object for it, and a new table, etc...)

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
```diff
@@ -0,0 +1,42 @@
...
+@Table(name="issues")
+@EqualsAndHashCode(callSuper = true)
+@NoArgsConstructor
+public class Issue extends AbstractIssue {
+
+	@NotNull	
+	@Column(name="title", length=128)
```
JWGmeligMeyling: I agree that the title should not be too big. But constraints also force us to check them. Maybe we should just leave the default value of `255` characters here, the storage penalty is not too bad anyway.

Also, I think it was a mistake that PR's cannot have names right now. Therefore I think this field should move to `AbstractIssue`.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
```diff
@@ -0,0 +1,42 @@
...
+
+
+@Data
+@Entity
+@Table(name="issues")
+@EqualsAndHashCode(callSuper = true)
+@NoArgsConstructor
+public class Issue extends AbstractIssue {
+
+	@NotNull
```
JWGmeligMeyling: `@NotEmpty` would make more sense to me

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
```diff
@@ -0,0 +1,42 @@
...
+@NoArgsConstructor
+public class Issue extends AbstractIssue {
+
+	@NotNull	
+	@Column(name="title", length=128)
+	private String title;
+	
+	@NotNull
+	@Column(name="description", length=2048)
+	private String description; // Move this to AbstractIssue sometime?
```
JWGmeligMeyling: I am thinking of making this field a `Text` field (without length limits). See the `BuildResult` on how to do that.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
```diff
@@ -0,0 +1,42 @@
...
+@Table(name="issues")
+@EqualsAndHashCode(callSuper = true)
+@NoArgsConstructor
+public class Issue extends AbstractIssue {
+
+	@NotNull	
+	@Column(name="title", length=128)
+	private String title;
+	
+	@NotNull
```
JWGmeligMeyling: @NotEmpty would make more sense to me

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
```diff
@@ -0,0 +1,42 @@
...
+	@Column(name="description", length=2048)
+	private String description; // Move this to AbstractIssue sometime?
+	
+	@Column(name="assignee")
+	@ManyToOne(optional = true, cascade=CascadeType.DETACH)
```
JWGmeligMeyling: Why specifically cascade `DETACH`? Looks a bit random to me :wink: I think we can quite safely cascade all other operations as well.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
AZWN: Just a random though: what happens to existing pull-requests when we add a `@NonEmpty` description field?

JWGmeligMeyling:
Good start! Of all the introduced fields, I think they are best stored in `AbstractIssue` itself, even though Pull request do not currently support it, and changing pull requests may not be in the scope of this issue.

AZWN:
Nice to hear :smile: What would you like as `Issue` specific features instead?

JWGmeligMeyling:
I am looking for differences between Github issues and pull requests, and the only difference I seem to find is that one has code and the other doesn't :wink:

So maybe, until we have some good ideas, this class can just be empty! (But we do need it though!)

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
JWGmeligMeyling: I think these checks are only performed on persist time. They are done by the separate `hibernate-validator` anyway, and not part of the JPA core. (For that, you should use `@Column(nullable = false)`). But if we start storing the name, we should create one for the pull requests yes. That makes me question how the names are currently being generated. To answer your question, to our database migration, we simply add a query that sets the title field for pull requests, based on a sensible schema (`SET title = CONCAT("Pull Request", issueNumber, ...)`.

JWGmeligMeyling: I have answered your question for `title` instead of `description`. Now I think of it, I think a description should be both `nullable` and allow the empty string.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/Issue.java
AZWN: Wouldn't cascade `All` mean that e.g. when you delete a user, all issues assigned to that user will be deleted too?

JWGmeligMeyling: That could indeed be the case, you may however need `orphanRemoval = true` for that. But lets go for the safe route and add PERSIST and MERGE manually ;)

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Issues.java
```diff
@@ -0,0 +1,29 @@
...
+public class Issues extends Controller<Issue> {
+
+	@Inject
+	public Issues(EntityManager em) {
+		super(em);
+	}
+	
+	@Transactional
+	public List<Issue> findIssues(final RepositoryEntity repo){
+		return query().from(issue)
```
JWGmeligMeyling: This can potentially be replaced with:
```java
@OneToMany(mappedBy = "repository")
List<Issue> issues;
```
on `AbstractRepository`.

But maybe we will never really access all issues at once, and should we have separate queries for open and closed issues.

src/main/resources/changelog.xml
```diff
@@ -971,5 +971,41 @@
 			</column>
 		</addColumn>
 	</changeSet>
+	
+	<changeSet id="add_issues" author="Aron Zwaan">
+		<createTable tableName="repository_issues">
+			<column name="issue_id" type="bigint" autoIncrement="true">
+				<constraints primaryKey="true" nullable="false" />
```
JWGmeligMeyling: Issue numbers are shared with pull requests, and start counting at 1 for each repository. Therefore the primary key is a compound key between `issue_id` and `repository_id`. See also the migration for pull requests.

src/main/resources/changelog.xml
```diff
@@ -971,5 +971,41 @@
 			</column>
 		</addColumn>
 	</changeSet>
+	
+	<changeSet id="add_issues" author="Aron Zwaan">
+		<createTable tableName="repository_issues">
+			<column name="issue_id" type="bigint" autoIncrement="true">
```
JWGmeligMeyling: Then `autoincrement=true` is not necessary as well, because issue_ids are programatically generated.

JWGmeligMeyling:
👍 Tests pass again :)

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Issues.java
AZWN: I was planning to implement tqueries for open/closed/issues per user net haha :smile:

JWGmeligMeyling:
👍

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Issues.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class Issues extends Controller<Issue> {
+
+	@Inject
+	public Issues(EntityManager em) {
+		super(em);
+	}
+	
+	@Transactional
+	public List<Issue> findIssues(final RepositoryEntity repo, User user){
```
JWGmeligMeyling: This should probably be `findAssignedIssues`?

JWGmeligMeyling:
Looks good so far! :+1:

AZWN:
Whoops, it seems travis handles line breaks somewhat different :?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
```diff
@@ -0,0 +1,200 @@
...
+		RepositoryApi repositoryApi = getRepositoryApi(repositoryEntity);
+		RepositoryModel repository = repositoryApi.getRepositoryModel();
+		
+		Issue issue = issues.findIssueById(getRepositoryEntity(), issueId).get(0);
+		
+		issue.setTitle(title);
+		issue.setDescription(description);
+		
+		User assignee = users.findByNetId(assigneeNetID);
+		checkCollaborator(assignee);
```
JWGmeligMeyling: If we can never assign an issue to a non collaborator, the list will be empty anyway. I think this security check is redundant. You get what you ask for, nothing :wink;

AZWN: They always taught me to never rely on the frontend for security issues. In theory, I could make a request assigning an issue to someone not collaborating. Thats why I inserted this check....

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
```diff
@@ -0,0 +1,200 @@
...
+		
+		issue.setTitle(title);
+		issue.setDescription(description);
+		issue.setOpen(true);
+		
+		User assignee = users.findByNetId(assigneeNetID);
+		checkCollaborator(assignee);
+		
+		issue.setAssignee(assignee);
+		issue.setTimestamp(new Date());
```
JWGmeligMeyling: I believe this field has the `@Creation` annotation, so this assignment is not needed, and done by Hibernate under the hood.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
```diff
@@ -0,0 +1,200 @@
...
+
+	@POST
+	@Transactional
+	@Path("/issue/{issueId}")
+	public Response updateIssue(@Context HttpServletRequest request, 
+			@PathParam("issueId") long issueId,
+			@FormParam("title") String title,
+			@FormParam("description") String description,
+			@FormParam("assignee") String assigneeNetID,
+			@FormParam("status") String status) throws IOException {
```
JWGmeligMeyling: As status only accept `open` and `closed` it's maybe more descriptive to make this a Java enum type. JAX-RS accepts enums as parameters. By convention, you may have to switch to full uppercase words though :)

JWGmeligMeyling:
I have almost no comments. Lets keep this as-is. You want me to look into adding the comments to the abstracted level and push the result to your branch, or do you see any further problems with that?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
JWGmeligMeyling: I read the function incorrect, updated my comment

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
AZWN: You were right :smile: I removed it.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
AZWN: I understand your point, but would you want is implemented in `AbstractIssue`, or just in the test? Since changing it in `AbstractIssue` would also have impact on `PullRequest`s....

JWGmeligMeyling:
What is the status of this branch? :wink: Still writing tests?

AZWN:
Yes, indeed writing tests and fixing styling issues

AZWN:
Ok, I think this feature is quite ready now :smile:

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/AbstractIssue.java
```diff
@@ -71,6 +77,18 @@
 	@Temporal(TemporalType.TIMESTAMP)
 	private Date closed;
 
+	// Left away @NotEmpty for now, since existing tests should be updated for it to work
```
JWGmeligMeyling: Can we bring this annotation back? As there is currently no other mechanism to prevent empty title names, which I think is relatively unpreferable.

The current naming schema for pull requests is `Pull Request #${pullRequest.issueId}: ${pullRequest.branchName}` (and hardcoded in the Freemarker templates). Perhaps it is an idea to persist this automatically generated name in the `PullRequestsBackend` and also use the `${pullRequest.title}` field from Freemarker from now on. (Until we also allow the pull request title to be set manually).

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
JWGmeligMeyling: It seems it's still included in this diff though?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
```diff
@@ -0,0 +1,231 @@
...
+		parameters.put("issue", issue);
+		
+		List<Locale> locales = Collections.list(request.getLocales());
+		return display(templateEngine.process("courses/assignments/group-issue-view.ftl", locales, parameters));
+	}
+
+	@POST
+	@Transactional
+	@Path("/issue/{issueId}/edit")
+	public Response updateIssue(@Context HttpServletRequest request,
```
JWGmeligMeyling: You can also add `@Context` parameters as fields to your resource class. (They don't require another annotation such as `@Inject`, because they are injected by JAX-RS (and CDI if you would use that) and not Guice). Passing the HttpServletRequest as field rather than method parameter makes it easier to call this methods from tests / other parts of the code.

*No need to change this now*

src/test/java/nl/tudelft/ewi/devhub/webtests/IssuesTest.java
```diff
@@ -0,0 +1,162 @@
+package nl.tudelft.ewi.devhub.webtests;
+
+import javax.inject.Inject;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.openqa.selenium.By;
+
+import com.mchange.util.AssertException;
```
JWGmeligMeyling: Unused import I suppose :wink:

src/test/java/nl/tudelft/ewi/devhub/webtests/views/IssueView.java
```diff
@@ -0,0 +1,24 @@
+package nl.tudelft.ewi.devhub.webtests.views;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.Select;
+
+public abstract class IssueView extends AuthenticatedView {
```
JWGmeligMeyling: What is exactly the philosophy behind this class? Can't this be an interface that views may implement? I am afraid that with this view we potentially hit an diamond problem with the view structure.

JWGmeligMeyling:
Great work! The `IssueBackend` seems to be not necessary right now, but as it is consistent with pull requests, I would keep it as is. I am going to spend this weekend merging PR's and redeploying I think. There is not a real need for changes, but have a look at my comments :wink:

AZWN:
Haha thnx :)
I've read your comments. Personally, I like to try and make things as good as possible. This week I don't have time left, but I would like to fix the small issues :D

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
JWGmeligMeyling: I believe you can even use a `Boolean` here, which has the same type as the field on `PullRequest`, where `=true` is `=true`, `false` is `false` and undefined is `null`. If you create the enum instead  I would specifically make one for the resource as this is API-wise convenience, and changing it in the entity as well will be very complicated, as you have to change the field, its Hibernate mapping, and probably some sophisticated database changelog as well :wink:

JWGmeligMeyling:
Ok I'll see whenever your changes arrive

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
AZWN: Ow, I probably resetted that commit again :confused:

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/AbstractIssue.java
AZWN: I have some small issues implementing this for pull-requests. The point is that the title must be present at creation time, but the id is not generated then. This gives a typical "Chicken or the egg" problem. We could place some dummy content, and update is directly after persisting, but that seems a bit hackish to me. Do you have an idea about how to fix that cleanly?

AZWN:
I've adressed some of the issues you pointed out.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/AbstractIssue.java
JWGmeligMeyling: Ah yes... That is indeed a problem. You could go through the hassle of writing your own Hibernate value **generator**. But I have a work around for now:

```java
pullRequest.setTitle(branchName);
pullRequests.persist(pullRequest);
pullrequest.setTitle(computeTitle...);
```

src/main/resources/templates/courses/assignments/group-pulls.ftl
```diff
@@ -37,7 +37,7 @@
                               </div>
                               <div class="pull-right"><i class="glyphicon glyphicon-comment"></i> ${pullRequest.comments?size} </div>
                             </span>
-                            <div class="comment">Pull Request #${pullRequest.issueId}: ${pullRequest.branchName}</div>
+                            <div class="comment">#${pullRequest.title}</div>
```
JWGmeligMeyling: Hash tag?

JWGmeligMeyling:
Looks very promising on the first sight! Thanks for all the work. Will checkout the branch tonight.

src/main/resources/templates/courses/assignments/group-pulls.ftl
```diff
@@ -37,7 +37,7 @@
                               </div>
                               <div class="pull-right"><i class="glyphicon glyphicon-comment"></i> ${pullRequest.comments?size} </div>
                             </span>
-                            <div class="comment">Pull Request #${pullRequest.issueId}: ${pullRequest.branchName}</div>
+                            <div class="comment">${pullRequest.title}</div>
```
JWGmeligMeyling: We need a migration

```sql
update pull_requests set title = concat('Pull Request #', issue_id, ':', branch_name)  where title is null;
```

for this.

JWGmeligMeyling: The line https://github.com/devhub-tud/devhub/pull/384/files#diff-e3b9c5a72b297c49ed8754b7d71cda8aR74 should be updated as well.

AZWN: I was already afraid for that migrations, that's why it was commented out for so long. Do you want that migration in the changelog.xml?

AZWN: Considered the freemarker, I'll fix that asap.

PR #385, opened 2016-05-24
Large commit warning test. Fixes #362

### LINKED ISSUE #362: Create a unit test for LargeCommitWarningGenerator
Create a unit test for `LargeCommitWarningGenerator`, determine which dependencies to mock, which methods to stub and assess the testability of the `LargeCommitWarningGenerator`. Reflect on the code coverage improvement. Also reflect on the usefulness of the warning it produces.

Fastjur:
Issue #362

### DIFF
--- a/src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
@@ -1,7 +1,6 @@
 package nl.tudelft.ewi.devhub.server.backend.warnings;
 
 import com.google.common.collect.Sets;
-import nl.tudelft.ewi.devhub.server.database.Configurable;
 import nl.tudelft.ewi.devhub.server.database.entities.Commit;
 import nl.tudelft.ewi.devhub.server.database.entities.RepositoryEntity;
 import nl.tudelft.ewi.devhub.server.database.entities.warnings.LargeCommitWarning;
@@ -16,15 +15,17 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
-import org.mockito.stubbing.OngoingStubbing;
 
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.anyObject;
+import static org.mockito.Matchers.anyString;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -32,12 +33,8 @@ import static org.mockito.Mockito.when;
 public class LargeCommitWarningGeneratorTest {
 
     private LargeCommitWarningGenerator generator;
-    private LargeCommitWarning warning;
+    private Set<LargeCommitWarning> testEquals;
 
-    private static final int MAX_AMOUNT_OF_FILES = 10;
-    private static final int MAX_AMOUNT_OF_LINES_TOUCHED = 500;
-    private static final String MAX_FILES_PROPERTY = "warnings.max-touched-files";
-    private static final String MAX_LINE_TOUCHED_PROPERTY = "warnings.max-line-edits";
     private static final String REPOSITORY_NAME = "John Cena";
     private static final String COMMIT_ID = "1";
 
@@ -50,38 +47,55 @@ public class LargeCommitWarningGeneratorTest {
     @Mock private CommitApi commitApi;
     @Mock private DiffModel diffModel;
     @Mock private RepositoryApi repositoryApi;
+    @Mock private Collection collection;
+    @Mock private Stream<AbstractDiffModel.DiffFile<AbstractDiffModel.DiffContext<AbstractDiffModel.DiffLine>>> stream;
+    @Mock private Stream<Object> objectStream;
 
     @Before
     public void setUp() {
         generator = new LargeCommitWarningGenerator(repositoriesApi);
-        warning = new LargeCommitWarning();
+
+        /* Build the mocked commit */
         when(commit.getRepository()).thenReturn(configurable);
         when(commit.getRepository().getRepositoryName()).thenReturn(REPOSITORY_NAME);
         when(commit.getCommitId()).thenReturn(COMMIT_ID);
+        when(commit.isMerge()).thenReturn(false);
+
         when(abstractCommitWarningGenerator.getRepository(commit)).thenReturn(repositoryApi);
         when(abstractCommitWarningGenerator.getGitCommit(commit)).thenReturn(commitApi);
+
+        /* Build the mocked repo */
+        when(repositoriesApi.getRepository(REPOSITORY_NAME)).thenReturn(repositoryApi);
         when(repositoryApi.getCommit(COMMIT_ID)).thenReturn(commitApi);
         when(commitApi.diff()).thenReturn(diffModel);
         when(diffModel.getDiffs()).thenReturn(diffs);
-        when(repositoriesApi.getRepository(REPOSITORY_NAME)).thenReturn(repositoryApi);
-        when(commit.isMerge()).thenReturn(false);
+
+        /* Used to prevent NullPointerExceptions in tooManyLineChanges */
+        when(diffs.stream()).thenReturn(stream);
+        when(stream.filter(anyObject())).thenReturn(stream);
+        when(stream.flatMap(anyObject())).thenReturn(objectStream);
+        when(objectStream.flatMap(anyObject())).thenReturn(objectStream);
+        when(objectStream.filter(anyObject())).thenReturn(objectStream);
+
+        /* Set up a warning, used to test if the generator returns an equal object */
+        LargeCommitWarning warning = new LargeCommitWarning();
+        warning.setCommit(commit);
+        testEquals = Sets.newHashSet(warning);
     }
 
     @Test
     public void testIsMerge() {
         when(commit.isMerge()).thenReturn(true);
-        Set empty = generator.generateWarnings(commit, gitPush);
+        Set<LargeCommitWarning> empty = generator.generateWarnings(commit, gitPush);
         assertTrue(empty.isEmpty());
     }
 
     @Test
     public void testTooManyFiles() {
         when(diffs.size()).thenReturn(1);
-        when(configurable.getIntegerProperty(MAX_FILES_PROPERTY, MAX_AMOUNT_OF_FILES)).thenReturn(0);
+        when(configurable.getIntegerProperty(anyString(), anyString())).thenReturn(0);
 
-        warning.setCommit(commit);
-        Set testEquals = Sets.newHashSet(warning);
-        Set warnings = generator.generateWarnings(commit, gitPush);
+        Set<LargeCommitWarning> warnings = generator.generateWarnings(commit, gitPush);
 
         verify(diffs).size();
 
@@ -89,4 +103,17 @@ public class LargeCommitWarningGeneratorTest {
         assertEquals(1, warnings.size());
         assertEquals(testEquals, warnings);
     }
+
+    @Test
+    public void testTooManyLineChanges() {
+        when(objectStream.count()).thenReturn((long) 1);
+        when(configurable.getIntegerProperty(anyString(), anyString())).thenReturn(0);
+
+        Set<LargeCommitWarning> warnings = generator.generateWarnings(commit, gitPush);
+
+        assertFalse(warnings.isEmpty());
+        assertEquals(1, warnings.size());
+        assertEquals(testEquals, warnings);
+    }
+
 }
--- a/src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
@@ -9,10 +9,12 @@ import nl.tudelft.ewi.devhub.server.database.entities.PrivateRepository;
 import nl.tudelft.ewi.devhub.server.database.entities.warnings.LargeCommitWarning;
 import nl.tudelft.ewi.devhub.server.web.models.GitPush;
 import nl.tudelft.ewi.git.models.AbstractDiffModel;
+import nl.tudelft.ewi.git.models.ChangeType;
 import nl.tudelft.ewi.git.models.DiffModel;
 import nl.tudelft.ewi.git.web.api.CommitApi;
 import nl.tudelft.ewi.git.web.api.RepositoriesApi;
 import nl.tudelft.ewi.git.web.api.RepositoryApi;
+import org.apache.commons.io.output.CloseShieldOutputStream;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -47,7 +49,7 @@ public class LargeCommitWarningGeneratorTest {
     @Mock private GitPush gitPush;
 
     @Before
-    public void setUp() {
+    public void setUp() throws IOException {
         generator = new LargeCommitWarningGenerator(repositoriesApi);
 
         repository = new PrivateRepository();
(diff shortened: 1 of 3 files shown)


src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
```diff
@@ -0,0 +1,137 @@
...
+public class LargeCommitWarningGeneratorTest {
+
+    private LargeCommitWarningGenerator generator;
+    private Set<LargeCommitWarning> testEquals;
+
+    private static final String REPOSITORY_NAME = "John Cena";
+    private static final String COMMIT_ID = "1";
+
+    @Mock private RepositoriesApi repositoriesApi;
+    @Mock private Commit commit;
```
JWGmeligMeyling: `Commit` is a data class, usually it is better/easier to use an actual object instead of a mock there (`new Commit()`)

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
```diff
@@ -0,0 +1,137 @@
...
+    @Mock private RepositoryEntity configurable;
```
JWGmeligMeyling: The same counts for `AbstractRepository`, but as this is an `abstract` class, you should chose between a `GroupRepository` or a `PrivateRepository`. The latter is easier, as it allows you to set the configuration directly on the repository, rather than to have this inherited from the course.

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
```diff
@@ -0,0 +1,137 @@
...
+    @Mock private GitPush gitPush;
+    @Mock private List<AbstractDiffModel.DiffFile<AbstractDiffModel.DiffContext<AbstractDiffModel.DiffLine>>> diffs;
+    @Mock private AbstractCommitWarningGenerator abstractCommitWarningGenerator;
```
JWGmeligMeyling: Why exactly is this necessary, it seems unused in the rest of the tests?

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
```diff
@@ -0,0 +1,137 @@
...
+    @Mock private CommitApi commitApi;
+    @Mock private DiffModel diffModel;
+    @Mock private RepositoryApi repositoryApi;
+    @Mock private Collection collection;
+    @Mock private Stream<AbstractDiffModel.DiffFile<AbstractDiffModel.DiffContext<AbstractDiffModel.DiffLine>>> stream;
+    @Mock private Stream<Object> objectStream;
```
JWGmeligMeyling: This is a very deep stub, touching the very internals of the Java collection implementation. This is quite smelly. Would you have chosen to directly use an actual `DiffModel` object - or at least an actual `DiffFile` object, you could have set an `ArrayList` of `DiffLines` as content, without having to mock the collection and the streams.

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
```diff
@@ -0,0 +1,137 @@
...
+        when(repositoriesApi.getRepository(REPOSITORY_NAME)).thenReturn(repositoryApi);
+        when(repositoryApi.getCommit(COMMIT_ID)).thenReturn(commitApi);
+        when(commitApi.diff()).thenReturn(diffModel);
+        when(diffModel.getDiffs()).thenReturn(diffs);
+
+        /* Used to prevent NullPointerExceptions in tooManyLineChanges */
+        when(diffs.stream()).thenReturn(stream);
+        when(stream.filter(anyObject())).thenReturn(stream);
+        when(stream.flatMap(anyObject())).thenReturn(objectStream);
+        when(objectStream.flatMap(anyObject())).thenReturn(objectStream);
```
JWGmeligMeyling: It would avoid these complex stubs as well, and even makes the test a little bit more stable, as changes to the `LargeCommitWarningGenerator` that start to use the underlying `Collection` interface differently will not require new specific stubs.

src/test/java/nl/tudelft/ewi/devhub/webtests/views/GroupEnrollView.java
```diff
@@ -49,6 +49,12 @@ public GroupEnrollView setMemberField(int number, String studentNetId) {
         WebElement memberField = getDriver().findElement(By.name(fullName));
         memberField.clear();
         memberField.sendKeys(studentNetId);
+        try {
+            /* This is necesarry to allow the field to update */
+            Thread.sleep(250);
```
JWGmeligMeyling: I believe @AZWN pointed out another alternative to this :)

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Confirms that no warning is generated when the commit is deemed good.
+     */
+    @Test
+    public void testGoodCommit() {
+        Set<LargeCommitWarning> warnings = generator.generateWarnings(commit, gitPush);
+
+        assertTrue(warnings.isEmpty());
```
JWGmeligMeyling: [`assertThat(warnings, is(empty()))`](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#empty()) is easier to read and provides a better error message when the test fails.

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
```diff
@@ -0,0 +1,137 @@
...
+    public void testTooManyFiles() {
+        when(diffs.size()).thenReturn(1);
+        when(configurable.getIntegerProperty(anyString(), anyString())).thenReturn(0);
+
+        Set<LargeCommitWarning> warnings = generator.generateWarnings(commit, gitPush);
+
+        verify(diffs).size();
+
+        assertFalse(warnings.isEmpty());
+        assertEquals(1, warnings.size());
```
JWGmeligMeyling: You are basically testing `Collections` here :wink: If warnings has size 1, it can never be empty. The latter assertion makes more sense to me. The first one is redundant :)

JWGmeligMeyling:
Nice work! :+1: Not all my comments have to be addressed, but please have a look at using some actual objects here and there, as it should make the test less verbose, and more open to extension of the warning generator.

What is your general thought on the testability of the `LargeCommitWarning`? And do you find it a sensible warning?

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
Fastjur: Uh oh, I forgot to clean up. I'll check unused mocks and delete them

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
Fastjur: I was struggling with this one quite a bit. Because the LargeCommitWarningGenerator contains this code:
```
int count = (int) diffFiles.stream()
      .filter(file -> !file.isDeleted() && fileViolation(file.getNewPath()))
      .flatMap(diffFile -> diffFile.getContexts().stream())
      .flatMap(context -> context.getLines().stream())
      .filter(line -> line.isAdded() || line.isRemoved())
      .count();
```

The only way to prevent NPE's was by actually stubbing the streams. Have you got any tips how to do this properly?

src/test/java/nl/tudelft/ewi/devhub/webtests/views/GroupEnrollView.java
Fastjur: Yes indeed, I'll try to implement this.

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
Fastjur: I thought that generally, one should Mock a classes dependencies, to prevent tests failing due to outside reasons (dependencies, like commit). Why does it make a difference that Commit is a Data class?

I'll update it!

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
JWGmeligMeyling: The `NullPointerException` is likely caused because it is deduced from the `newLineNumber` and `oldLineNumber` fields, in order to save space. Namely: if the line is added, it was not present in the old file, so the `oldLineNumber` is `null`. If the line was removed, it is not present in the new file, and the `newLineNumber` is `null`. If the line was present in both the new and old file, it can only be modified.

The implementation of `isAdded()` and `isRemoved()` uses these fields, and expect at least one of them to be not null. We do not add booleans / an enum value for `added`/`removed`/`modified` as this would require a lot of unnecessary space in the response (and consider it is sent over the wire for every single line in the diff).

As far as creating the diff object

1) Programatically creating a diff file
```java
		DiffLine diffLine = new DiffLine();
		diffLine.setNewLineNumber(1);
		diffLine.setContent("Hiephoi");
		
		DiffContext<DiffLine> diffContext = new DiffContext<>();
		diffContext.setLines(Lists.newArrayList(diffLine));
		DiffFile<DiffContext<DiffLine>> diffFile = new DiffFile<>();
		diffFile.setContexts(Lists.newArrayList(diffContext));
		DiffModel diffModel = new DiffModel();
		diffModel.setDiffs(Lists.newArrayList(diffFile));
```

2)
You can also create an example diff you need for your test, and convert it to JSON (if you are familiar with that). You can then store it as a JSON file under resources, and load it using the ObjectMapper:

```
ObjectMapper objectMapper = new ObjectMapper();
DiffModel diffModel = objectMapper.readValue(
   LargeCommitWarningGeneratorTest.class.getResourceAsStream("/my-diff.json"),
   DiffModel.class
);

File: src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java (hunk 1 above)
Comment (JWGmeligMeyling) [inline src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java]: The reason is that you *mock* a *dependency* because you want to *control* its *behaviour*. Data classes have no *behaviour*, but only contain some fields. They also don't have any object dependencies that would make their instantiation difficult. You already have the *controllability* you want, as you can just set its fields through their setters.

File: src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java (hunk 4 above)
Comment (Fastjur) [inline src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java]: Oh right, that makes sense, thanks!

Does this not contradict the mocking principles though? As failures may arise in DiffLine, or DiffContext. Causing the tests for LargeCommitWarningGenerator to fail, even though it is not a fault in this class.

File: src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java (hunk 1 above)
Comment (Fastjur) [inline src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java]: Right, my bad!

Comment (JWGmeligMeyling) [inline src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java]: Only difficulty is that `isMerge()` probably checks whether the commit has multiple parents, so you can't simply set a boolean to true, but have to set the collection of parents. While obviously the commit cannot point to itself, for the test it doesn't really matter, so you can be lazy here: `commit.setParents(Lists.newArrayList(commit, commit))`

Comment (Fastjur) [inline src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java]: Yes that is what I did. I just give it an empty list, or a list with multiple commits

File: src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java (hunk 4 above)
Comment (JWGmeligMeyling) [inline src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java]: During the Software Engineeringing principles course you may hear that data classes should be avoided any way. I do however think for a structure like this, it is hardly avoidable. Personally, I tend to like them very much for Object Relational Mapping and JSON/XML (de)serialization.

But yes, if there is a bug in our data classes, it's all around. Luckily they don't have so much logic, and most of it is auto generated :wink:

File: src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java (hunk 1 above)
Comment (JWGmeligMeyling) [inline src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java]: Also, data objects tend to be immutable (this is not completely the case for database entities which are being observed for changes though). Because they are immutable, it is also unlikely that you would want to verify an invocation on the object, the second purpose of mocking :)

File: src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java (hunk 4 above)
Comment (Fastjur) [inline src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java]: Problem with this is though, that that json file will actually have to be a file with too many lines. If the threshold of 'too many lines' ever gets changed, one would also have to change this resource to keep the test valid

File: src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java [hunk 9]
```diff
@@ -0,0 +1,131 @@
...
+    private static final String COMMIT_ID = "1";
+
+    @Mock private RepositoriesApi repositoriesApi;
+    @Mock private RepositoryApi repositoryApi;
+    @Mock private CommitApi commitApi;
+    @Mock private GitPush gitPush;
+
+    @Before
+    public void setUp() throws IOException {
+        generator = new LargeCommitWarningGenerator(repositoriesApi);
```
JWGmeligMeyling: You can by the way also declare this as a  field and use the `@InjectMocks` annotation on it :)

Fastjur:
Addressing your previous questions:

What is your general thought on the testability of the `LargeCommitWarning`?  
Testability was OK, had some trouble setting up the diffModel though, mainly due to my own incompetence. It was not hard to setup the actual test cases though.

And do you find it a sensible warning?  
This really depends on the situation, it is, however, convention to keep your commits small, as this improves readability and the general structure of the repo will be less messy. So yes, I do think this is a sensible warning

src/test/java/nl/tudelft/ewi/devhub/server/backend/warnings/LargeCommitWarningGeneratorTest.java
```diff
@@ -114,6 +110,12 @@ public void testTooManyFiles() {
      */
     @Test
     public void testTooManyLineChanges() {
+        List<AbstractDiffModel.DiffLine> diffLines = Lists.newArrayList();
+        for (int i = 0; i < MAX_AMOUNT_OF_LINES_TOUCHED + 2; i++) {
+            AbstractDiffModel.DiffLine diffLine = new AbstractDiffModel.DiffLine(2*i, 2*i+1, String.valueOf(i));
```
JWGmeligMeyling: As you set both the `newLineNumber` and the `oldLineNumber` these lines are assumed to be *unchanged* (as a modified line would be split up in to the removal of the old version, and the addition of the new version of the line). As they are unchanged, they are not counted for the warning (by design). You therefore have to adjust the test so that these lines are either *added* to or *removed* from the file. You can do this by setting one of the counts to `null`.

JWGmeligMeyling:
> Testability was OK, had some trouble setting up the diffModel though, mainly due to my own incompetence. It was not hard to setup the actual test cases though.

The `DiffModel` is indeed hard to set up. I think this also comes because its structure relies quite heavily on the use of generics, in order to use mostly the same structure for `DiffBlameModels` (which are `DiffModels` with authorship information added per line).

> This really depends on the situation, it is, however, convention to keep your commits small, as this improves readability and the general structure of the repo will be less messy. So yes, I do think this is a sensible warning.

Yes, there will be a lot of false positives. Thats also one reason why the max size is configurable. I am however not sure what the currently configured threshold is 👼

Fastjur:
> Yes, there will be a lot of false positives. Thats also one reason why the max size is configurable. I am however not sure what the currently configured threshold is :angel:

False positives are actually okay. Haven't had them occur yet, and committing something with 500+ lines *should* feel a bit off anyway. The current threshold is 10 files and 500 lines by the way, seems reasonable to me.

JWGmeligMeyling:
@Fastjur could you perhaps bring this branch up to date with master?

Fastjur:
@JWGmeligMeyling Yep sorry, it is up to date now.

JWGmeligMeyling:
Thanks! 👍

PR #391, opened 2016-06-03
[WIP] add markdown support for comments, assignments and mardown files

### LINKED ISSUE #149: Markdown for comments, assignments, md files in repo

DouweKoopmans:
fixes #149

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/util/MarkDownParser.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/util/MarkDownParser.java
@@ -6,6 +6,7 @@ import lombok.Synchronized;
 import org.parboiled.errors.ParserRuntimeException;
 import org.pegdown.PegDownProcessor;
 
+import javax.inject.Inject;
 import javax.validation.constraints.NotNull;
 
 /**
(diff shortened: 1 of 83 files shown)


DouweKoopmans:
don't know why coveralls says the coverage went up, i haven't written any tests yet

JWGmeligMeyling:
I have added one commit to the PR. This commit addresses issues related to static state. In Devhub we provide dependecies through dependency injection. Static access is considered a bad practise. This commit
addresses that problem, and provides the `PegDownParser` and `MarkDownParser` by DI.

Concurrent access to the `PegDownParser` is prevented by binding it to the `RequestScope`.
No synchronized access is needed, so we keep our concurrency high.

As the `TemplateEngine` was accessed as a singleton, and it depends on the `PegDownParser`,
access to the `TemplateEngine` had to be transformed to `RequestScoped`, which is done
by using Guice `Providers`.

src/test/java/nl/tudelft/ewi/devhub/server/util/MarkDownParserTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+		assertEquals(expectedHtml, markDownParser.markdownToHtml(testMd));
+	}
+
+	@Test
+	public void testTimeout() {
+		markDownParser = new MarkDownParser(new PegDownProcessor(2000l));
+		String testMd = "*Hello World!*";
+		String expectedHtml = "<p><em>Hello World!</em></p>";
+		assertEquals(expectedHtml, markDownParser.markdownToHtml(testMd));
```
JWGmeligMeyling: This test does not test what it should. Furthermore, the test suite is very limited. There is no realistic good-weather test (a bigger chunk of data, with multiple lines), and there is no test for the potential cross-site scripting problems. These tests should be added.

src/main/java/nl/tudelft/ewi/devhub/server/util/MarkDownParser.java
```diff
@@ -0,0 +1,43 @@
...
+import org.parboiled.errors.ParserRuntimeException;
+import org.pegdown.PegDownProcessor;
+
+import javax.inject.Inject;
+import javax.validation.constraints.NotNull;
+
+/**
+ * Created by Douwe Koopmans on 1-6-16.
+ */
+// TODO: 2-6-16 add preview panel when writing markdown
```
JWGmeligMeyling: Please don't leave TODO's in the code base. These should be added as issue on the board.

JWGmeligMeyling:
This implementation suffers from the problem that Pegdown also escapes HTML code, but only within code blocks. Therefore code blocks are now escaped twice, resulting in strange output. We should however run the request over an escaper, as Pegdown only escapes code blocks, and this would make us vulnerable to XSS.

See: https://github.com/sirthias/pegdown/issues/236

JWGmeligMeyling:
The ideal way to have objects that write to the Freemarker output is to use a directive.
This commit implements such a directive on the existing Markdown parser object.
With this change, `[#noescape]` tags around the Markdown blocks are no longer required.
Furthermore, we do not need the `BeansWrapper` anymore.

The current syntax is:

```freemarker
[@MarkDownParser message=comment.content/]
```

Now lets keep an eye on the duplicate escaping issue. Imho no show stopper, as most people will put Java code in the code blocks in our use case anyway.

DouweKoopmans:
unfortunately quotation marks are being escaped, so this creates weird output when working with java strings.

JWGmeligMeyling:
Ah thats a shame... Maybe we should fix Pegdown?

PR #401, opened 2016-06-13
Add end-to-end tests for deletion of ahead branches

Fastjur:
Including tests

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
@@ -153,12 +153,16 @@ public abstract class AbstractProjectResource<RepoType extends RepositoryEntity>
 	 * Generates the parameters for the branch overview.
 	 * @param branchName Name of the branch
 	 * @param page Page number to display
-	 * @param branchDeletion Boolean value to indicate successfull branch deletion. <b>TRUE</b>
-	 *                          means success, <b>FALSE</b> means failure. <b>NULL</b> means no
-	 *                          branch deletion occurred.
+	 * @param branchDeletionStatus String representing the status of a branch deletion.
+	 *                <ul>
+	 *                  <li><b>success</b> means that the branch was successfully deleted</li>
+	 *                  <li><b>error</b> means that an error occurred
+	 *                  <li><b>confirm</b> means that the branch deletion requires confirmation</li>
+	 *                </ul>
      * @return A map containing the response parameters.
      */
-        protected Map<String, Object> getBranchOverviewParameters(String branchName, int page, Boolean branchDeletion) {
+        protected Map<String, Object> getBranchOverviewParameters(String branchName, int page,
+																  String branchDeletionStatus) {
             RepositoryEntity repositoryEntity = getRepositoryEntity();
 		RepositoryApi repositoryApi = repositoriesApi.getRepository(repositoryEntity.getRepositoryName());
 		Map<String, Object> parameters = getBaseParameters();
@@ -188,8 +192,8 @@ public abstract class AbstractProjectResource<RepoType extends RepositoryEntity>
 			} else throw e;
 		}
 
-		if (branchDeletion != null) {
-			parameters.put("deleteSuccessful", branchDeletion);
+		if (branchDeletionStatus != null) {
+			parameters.put("deleteStatus", branchDeletionStatus);
 		}
 
 		return parameters;
@@ -600,9 +604,11 @@ public abstract class AbstractProjectResource<RepoType extends RepositoryEntity>
 	@POST
 	@Path("/branches/delete")
 	@Transactional
-	public Response deleteBehindBranch(@Context HttpServletRequest request,
-									   @FormParam("branchDeleteName") String branchDeleteName,
-									   @QueryParam("fatal") String fatal)
+	public Response deleteBranch(@Context HttpServletRequest request,
+                                 @FormParam("branchDeleteName") String branchDeleteName,
+                                 @FormParam("branchDeleteNameConfirmation")
+                                             String branchDeleteNameConfirmation,
+                                 @QueryParam("fatal") String fatal)
 			throws IOException, ApiError {
 
 		RepositoryEntity repositoryEntity = getRepositoryEntity();
@@ -617,12 +623,36 @@ public abstract class AbstractProjectResource<RepoType extends RepositoryEntity>
 			BranchModel branchModel = branchApi.get();
 
 			if (!branchModel.isAhead()) {
+                // Branch is not ahead and can be safely removed
 				branchApi.deleteBranch();
-				parameters = getBranchOverviewParameters("master", 1, true);
+				parameters = getBranchOverviewParameters("master", 1, "success");
 			} else {
-				parameters = getBranchOverviewParameters("master", 1, false);
+                // Branch is ahead
+                if (branchDeleteNameConfirmation != null) {
+                    if (branchDeleteName.endsWith(branchDeleteNameConfirmation)
+                            && !branchDeleteName.equals("refs/heads/master")) {
+                        // Confirmation is correct and branch is not master, delete ahead branch.
+                        branchApi = repositoryApi.getBranch(branchDeleteName);
+                        branchModel = branchApi.get();
+                        branchApi.deleteBranch();
+                        parameters = getBranchOverviewParameters("master", 1, "success");
+                        parameters.put("deletedBranchSimpleName", branchModel.getSimpleName());
+                    } else {
+                        // Wrong confirmation, prompt user again
+                        parameters = getBranchOverviewParameters("master", 1, "confirmAgain");
+                        parameters.put("aheadBranchName", branchModel.getName());
+                        parameters.put("aheadBranchSimpleName", branchModel.getSimpleName());
+                    }
+                } else {
+                    // Need confirmation to delete this branch
+                    parameters = getBranchOverviewParameters("master", 1, "confirm");
+                    parameters.put("aheadBranchName", branchModel.getName());
+                    parameters.put("aheadBranchSimpleName", branchModel.getSimpleName());
+                }
 			}
-        }
+        } else {
+			parameters = getBranchOverviewParameters("master", 1, "error");
+		}
 
         return display(templateEngine.process("project-view.ftl", locales, parameters));
 	}
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
@@ -616,7 +616,7 @@ public abstract class AbstractProjectResource<RepoType extends RepositoryEntity>
 
         List<Locale> locales = Collections.list(request.getLocales());
 
-        Map<String, Object> parameters = getBaseParameters();
+        Map<String, Object> parameters;
 
         if (branchDeleteName != null) {
             BranchApi branchApi = repositoryApi.getBranch(branchDeleteName);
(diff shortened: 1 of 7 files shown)


Fastjur:
WTF. Why does this have emoji commits in it again?

Fastjur:
There we go, nice and clean. Had to channel my git-fu and smack my PC a little but it worked.

Fastjur:
Coveralls is confused, but otherwise **ready to merge**.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
```diff
@@ -188,8 +192,8 @@ protected AbstractProjectResource(final TemplateEngine templateEngine,
 			} else throw e;
 		}
 
-		if (branchDeletion != null) {
-			parameters.put("deleteSuccessful", branchDeletion);
+		if (branchDeletionStatus != null) {
+			parameters.put("deleteStatus", branchDeletionStatus);
```
JWGmeligMeyling: I thought having a `Boolean` was preferable, and requires less boilerplate as well? :wink:

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
```diff
@@ -600,29 +604,56 @@ public Response showSettings(@Context HttpServletRequest request) throws IOExcep
...
+                        // Wrong confirmation, prompt user again
+                        parameters = getBranchOverviewParameters("master", 1, "confirmAgain");
+                        parameters.put("aheadBranchName", branchModel.getName());
+                        parameters.put("aheadBranchSimpleName", branchModel.getSimpleName());
+                    }
+                } else {
+                    // Need confirmation to delete this branch
+                    parameters = getBranchOverviewParameters("master", 1, "confirm");
+                    parameters.put("aheadBranchName", branchModel.getName());
+                    parameters.put("aheadBranchSimpleName", branchModel.getSimpleName());
```
JWGmeligMeyling: Oh this has gotten quite complicated, what is all of this doing?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
Fastjur: I more than 2 values to specify the type of alert that is shown. It is the easiest and simplest way of implementing a confirmation for deleting ahead branches

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
Fastjur: This is used to display the different types of branch deletion alerts. I will see if I can restructure this for clarity.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
```diff
@@ -600,29 +604,56 @@ public Response showSettings(@Context HttpServletRequest request) throws IOExcep
 	@POST
 	@Path("/branches/delete")
```
JWGmeligMeyling: What about `/branch/{branchName}/delete`?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
```diff
@@ -600,29 +604,56 @@ public Response showSettings(@Context HttpServletRequest request) throws IOExcep
...
 	@Transactional
-	public Response deleteBehindBranch(@Context HttpServletRequest request,
-									   @FormParam("branchDeleteName") String branchDeleteName,
-									   @QueryParam("fatal") String fatal)
+	public Response deleteBranch(@Context HttpServletRequest request,
+                                 @FormParam("branchDeleteName") String branchDeleteName,
```
JWGmeligMeyling: Then this becomes a `@PathParam`, and we're speaking of a `branchName`.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
```diff
@@ -600,29 +604,56 @@ public Response showSettings(@Context HttpServletRequest request) throws IOExcep
...
+                                 @QueryParam("fatal") String fatal)
 			throws IOException, ApiError {
 
 		RepositoryEntity repositoryEntity = getRepositoryEntity();
 		RepositoryApi repositoryApi = repositoriesApi.getRepository(repositoryEntity.getRepositoryName());
 
         List<Locale> locales = Collections.list(request.getLocales());
 
-        Map<String, Object> parameters = getBaseParameters();
+        Map<String, Object> parameters;
```
JWGmeligMeyling: Just use `getBaseParameters()` here, and add the `BranchModel` for the current branch under a variable. This gives you access in Freemarker to both the simple name and the actual name.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
```diff
@@ -600,29 +604,56 @@ public Response showSettings(@Context HttpServletRequest request) throws IOExcep
...
 			if (!branchModel.isAhead()) {
+                // Branch is not ahead and can be safely removed
 				branchApi.deleteBranch();
-				parameters = getBranchOverviewParameters("master", 1, true);
+				parameters = getBranchOverviewParameters("master", 1, "success");
 			} else {
-				parameters = getBranchOverviewParameters("master", 1, false);
+                // Branch is ahead
+                if (branchDeleteNameConfirmation != null) {
+					String branchSimpleName = branchDeleteName.split("refs/heads/")[1];
```
JWGmeligMeyling: Isn't this exactly what the simpleName used to be for? It looks very error prone to me as well. Technically it shouldn't have to be on `refs/heads`. Maybe we should just use the full branch name for the request anyway.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
Fastjur: Hmm that is a fair point. I will change this.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
```diff
@@ -600,29 +604,56 @@ public Response showSettings(@Context HttpServletRequest request) throws IOExcep
...
+                    if (branchSimpleName.equals(branchDeleteNameConfirmation)
+							&& !branchDeleteName.equals("refs/heads/master")) {
+                        // Confirmation is correct and branch is not master, delete ahead branch.
+                        branchApi = repositoryApi.getBranch(branchDeleteName);
+                        branchModel = branchApi.get();
+                        branchApi.deleteBranch();
+                        parameters = getBranchOverviewParameters("master", 1, "success");
+                        parameters.put("deletedBranchSimpleName", branchModel.getSimpleName());
```
JWGmeligMeyling: As `deletedBranchName` and `aheadBranchName` are comming from the same branch Model just use `branchModel` and use an other variable as flag.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
JWGmeligMeyling: Then it should be an enum, and not  strings :)

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
Fastjur: It's a bit smelly yeah, but the problem is, we want the confirmation the user types in to be the simple name (not refs/heads etc.). However if I check if the branchName ends with the confirmation, then typing in the last character of the branchName will succesfully delete it.

I could maybe split it up at every slash, and then get the last item in the array, which should be the branch name. This will however be prone to errors on branches including slashes. Hmm.

JWGmeligMeyling: You can just pass the `BranchModel` to Freemarker, and use the full name for URLs, and the simple name for display? No need to pass these two variables manually?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
Fastjur: Is there a notable advantage for this? Is it so we can send links to users possibly?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
Fastjur: Yeah I noticed this. This has been updated already.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
Fastjur: Hmm alright let me try that

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
JWGmeligMeyling: To conform REST-standards, it should have been `DELETE /branch/{branchName}`. But as `DELETE` is not supported in HTML4 forms, we have to use a `POST` alias request. I think `/branch/{branchName}` might be a little more consistent with the other requests, so it improves readability, but on the other hand, dealing with slashes in URL's is way more difficult, so we may not want to go through the hassle.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
JWGmeligMeyling: Still good to speak of a branch name instead of a branchDeleteName imho.

Fastjur: It is yeah. And thinking about it. If we use the pathparam we can include links to deletion in mails for stale branches etc. So I will change this :+1:

JWGmeligMeyling: hmm no that won't work, as it is a `POST` request and not a `GET`. (We could change that as well.) But I wouldn't change it just now, we have bigger issues to solve :wink: Have a look at replacing the "magic" strings with an enum though. JAX-RS is able to parse Enum values from strings, so it should be easy. And then you can make the logic less verbose/error prone/magic  and move some of the documentation to the enum  values instead .

Fastjur: Hmm okay will do this then. I changed it to an Enum. I'll see if I can add some clear documentation to it. The code size is already smaller.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectResource.java
Fastjur: It would have to be escaped using javascript I think. But maybe that is some optimization for later ;)

PR #405, opened 2016-06-15
Basic implementation of course view, fixes #270 

### LINKED ISSUE #270: Create course page
Right now the course page redirects to the current course edition. An overpage view should be implemented.
- [x] Course overview page
- [x] Ability to set up new course editions
- [ ] Ability to create course managers
- [ ] Seize own assistant permissions in favor of course managers

JWGmeligMeyling:
fixes #270


src/main/java/nl/tudelft/ewi/devhub/server/backend/CoursesBackend.java
```diff
@@ -33,7 +35,10 @@
 public class CoursesBackend {
 
     @Inject
-    private CourseEditions courses;
+    private CourseEditions courseEditions;
```
LiamClark: Nit picky, but how do you feel about changing the field injection to constructor injection?
But if we want to remove that let's do so in a different pr.

PR #413, opened 2016-06-24
Using chrome instead of firefox for webdriver.

avandeursen:
Firefox 47 does not play well with webdriver 2.53:
http://stackoverflow.com/questions/37693106/selenium-2-53-not-working-on-firefox-47

Switched to chrome instead of downgrading my firefox to 46. (you may have to (brew) install `chromedriver`)

For me two tests fail at the moment, both related to setting the state in the DeliveryReviewView.

> AssignmentReviewTest.testReviewForm:64 » NoSuchElement No value present
> AssignmentReviewTest.testSubmitReview:79 » NoSuchElement No value present

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -373,6 +373,18 @@
 		</dependency>
 		<dependency>
 			<groupId>org.seleniumhq.selenium</groupId>
+			<artifactId>selenium-chrome-driver</artifactId>
+			<version>${selenium.version}</version>
+			<scope>test</scope>
+			<exclusions>
+				<exclusion>
+					<artifactId>httpclient</artifactId>
+					<groupId>org.apache.httpcomponents</groupId>
+				</exclusion>
+			</exclusions>
+		</dependency>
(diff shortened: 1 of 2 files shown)


avandeursen:
And we should update the `.travis.yml` to download chromedriver too.

avandeursen:
The two tests failing for chrome have been fixed in #414, and now work for FF and Chrome.

avandeursen:
Getting travis to use chrome maybe a bit more involved than I thought.

And the tests [work well with Firefox 46](http://stackoverflow.com/a/38036477/165292), so I propose to close this PR for now.

JWGmeligMeyling:
Seems a bit complicated indeed. Linking to the related issue on the Travis tracker: https://github.com/travis-ci/travis-ci/issues/272#issuecomment-194370288 . Lets not remove the branch for future reference, maybe some developers want to use it in their personal configuration.

PR #414, opened 2016-06-24
setting status through select ignores whitespace

avandeursen:
In Chrome, tests failed because the setting a status included a comparison to the status getText(), which
included white space.

New test implementation ignores the whitespace, and now should work in both firefox and chrome.


JWGmeligMeyling:
Good catch! Will merge tomorrow :)

PR #415, opened 2016-06-25
Issue labels

JWGmeligMeyling:
@AZWN making a pr, makes intermediate review easier.


src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
```diff
@@ -132,6 +134,11 @@ public Response createIssue(
 		issue.setAssignee(assignee);
 		issue.setRepository(getRepositoryEntity());
 		
+		issue.getLabels().clear();
+		issue.getLabels().addAll(getRepositoryEntity().getLabels().stream().filter(
```
JWGmeligMeyling: `issue.setLabels`?

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
```diff
@@ -252,6 +258,25 @@ public Response addLabel(@FormParam("tag") String tag, @FormParam("color") Strin
 		);
 		return redirect(new URI(request.getRequestURI()).resolve("issues"));
 	}
+	
+	@POST
+	@Path("deleteLabel/{labelId}")
```
JWGmeligMeyling: According to REST conventions it is better to create the endpoints:

```java
@POST
@Path("label")
```

And:
```java
@DELETE
@Path("label/{labelId}")
```

Then in JQuery you would do:
```js
  return $.ajax({
    url: url,
    type: 'DELETE',
    success: callback,
    data: data,
    contentType: type
  });
```

See also http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api
And http://stepansuvorov.com/blog/2014/04/jquery-put-and-delete/

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/repository/AbstractProjectIssueResource.java
AZWN: `issue.setLabels` didn't quite work, but maybe a `merge` could fix that.

AZWN:
Darn, again some state preservation

AZWN:
@JWGmeligMeyling Can you look into that second failure? It seems it has to do with the `GroupCourseEnrollTest`. Locally though, it does pass.

AZWN:
Nevermind, after a restart it passed

src/test/java/nl/tudelft/ewi/devhub/webtests/IssuesTest.java
```diff
@@ -186,6 +198,15 @@ public void testAddRemoveLabel(){
 		assertEquals(0x00ccff, label.getColor());
 	}
 	
+	@After
+	public void deleteIssues(){
+		issues.findAllIssues(group.getRepository()).forEach(x -> {
+			Stream<? extends IssueComment> comments = issueComments.getMostRecentIssueComments(Lists.<GroupRepository>asList(group.getRepository(), new GroupRepository[0]),10);
+			comments.forEach(comment -> issueComments.delete(comment));
+			issues.delete(x);
```
JWGmeligMeyling: This can be done differently. Enable `orphanRemoval = true` and `CascadeType.DELETE` on `comments` in `AbstractIssue`. Then you can just do: `issues.findAllIssues(group.getRepository()).forEach(issues::delete)`.

AZWN: I understand your point :smile:, but that would let your tests influence the behaviour of the main code base, which is bad practise(I thought :wink: )

JWGmeligMeyling: True, but I think being able to delete an issue progremattically without having to delete its contents first is not so bad at all. But in practise issues won't be removed anyway ;)

avandeursen:
Quick question: I'm trying to build and test this, but the `testAddLabel` fails complaining that there is no `button.ms-choice` (in line 190 call to `selectLabels`).

All the other tests pass.

What's wrong in my setup?

AZWN:
I have introduced two new submodules. Weren't these updated?

src/test/java/nl/tudelft/ewi/devhub/webtests/IssuesTest.java
AZWN: That's true, I'll update it

avandeursen:
Ah indeed. `git submodule update --init` did the trick. All tests pass for me too now. Great.

PR #417, opened 2016-12-21
Migration to query dsl 4

LiamClark:
Querydsl moved to a new domain and has a new version out. 

The main issue is that 
.exists(); is gone in the newer version
The closest I could translate this is:
.fetchCount() > 0;

I opted for the fail fast FetchOne instead of FetchFirst, but this is debatable.

Further more map is no longer supported since they didn't like the usage of it and has become a group by.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/PrivateRepositories.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/PrivateRepositories.java
@@ -27,20 +27,20 @@ public class PrivateRepositories extends Controller<PrivateRepository> {
 		Preconditions.checkNotNull(username);
 		Preconditions.checkNotNull(repositoryTitle);
 
-		return ensureNotNull(query().from(privateRepository)
+		return ensureNotNull(query().selectFrom(privateRepository)
 				.where(privateRepository.owner.netId.equalsIgnoreCase(username)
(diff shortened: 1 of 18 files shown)


src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/PrivateRepositories.java
```diff
@@ -27,20 +27,20 @@ public PrivateRepository find(String username, String repositoryTitle) {
 		Preconditions.checkNotNull(username);
 		Preconditions.checkNotNull(repositoryTitle);
 
-		return ensureNotNull(query().from(privateRepository)
+		return ensureNotNull(query().selectFrom(privateRepository)
 				.where(privateRepository.owner.netId.equalsIgnoreCase(username)
 					.and(privateRepository.title.equalsIgnoreCase(repositoryTitle)))
-				.singleResult(privateRepository),
+				.fetchOne(),
```
JWGmeligMeyling: If fetchOne fails fast, then the ensureNotNull wrapper is unnecessary. But throughout Devhub we always propagate `EntityNotFound` exceptions, so if fetchOne throws something else, we should always catch and rethrow it, or may break other logic.

LiamClark: If there is no result it still returns null, leaving that logic valid, however it is stricter in the sense that for this query would there be two values, it throws a Not-unique exception.

PR #419, opened 2016-12-29
WIP Automatically fill in build rubric, fixes #277

### LINKED ISSUE #277: Automatic rubric for failing build

LiamClark:
Potential problems with this implementation:

When a student hands in a commit as a delivery, the commit can be in two states: Still building or finished.

The build of the final commit can change the outcome of the check.
So we check once when the delivery is made, and again when a commit that has been handed in finishes building. (Would there be any better way to combine these events?)

Also walking the commits like this is a query for every parent and might be quite slow.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/backend/ConsecutiveBuildFailureBackend.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/backend/ConsecutiveBuildFailureBackend.java
@@ -1,0 +1,118 @@
+package nl.tudelft.ewi.devhub.server.backend;
+
+import com.google.common.collect.Lists;
+import com.google.inject.Inject;
+import nl.tudelft.ewi.devhub.server.database.controllers.Commits;
+import nl.tudelft.ewi.devhub.server.database.controllers.Deliveries;
+import nl.tudelft.ewi.devhub.server.database.entities.Assignment;
+import nl.tudelft.ewi.devhub.server.database.entities.Commit;
+import nl.tudelft.ewi.devhub.server.database.entities.Delivery;
+import nl.tudelft.ewi.devhub.server.database.entities.Group;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Checks whether between the previous assignments delivery and the current deliver, or the start of time if it's the first assignment
+ * any commits through any branches that are reachable from the commit that has been handed in contained 4 or more consecutive build failures.
+ */
+public class ConsecutiveBuildFailureBackend {
+    public static final int MAX_CONSECUTIVE_BUILD_FAILURES = 4;
+    private Commits commits;
+    private Deliveries deliveries;
+
+    @Inject
+    public ConsecutiveBuildFailureBackend(Commits commits, Deliveries deliveries) {
+        this.commits = commits;
+        this.deliveries = deliveries;
+    }
+
+    public boolean hasConsecutiveBuildFailures(Group group, Delivery delivery) {
+        final List<List<Commit>> commitsToCheck = commitsToCheckForBuildFailures(group, delivery);
+        return consecutiveBuildFailures(commitsToCheck);
+    }
+
+    protected static boolean consecutiveBuildFailures(List<List<Commit>> commitsToCheck) {
+        return commitsToCheck.stream().anyMatch(ConsecutiveBuildFailureBackend::hasFourFailingCommits);
+    }
+
+
+    protected static boolean hasFourFailingCommits(List<Commit> commits) {
+        int consecutive = 0;
+
+        for (Commit commit : commits) {
+            if (commit.getBuildResult().hasFailed()) {
+                consecutive++;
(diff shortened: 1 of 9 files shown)


src/main/java/nl/tudelft/ewi/devhub/server/backend/ConsecutiveBuildFailureBackend.java
```diff
@@ -0,0 +1,118 @@
...
+            return commits.firstCommitInRepo(group.getRepository());
+        } else {
+            final Assignment previousAssignment = assignments.get(index - 1);
+            return deliveries.getLastDelivery(previousAssignment, group).map(Delivery::getCommit);
+        }
+    }
+
+    protected List<List<Commit>> commitsToCheckForBuildFailures(Group group, Delivery delivery) {
+        final Commit deliveryCommit = delivery.getCommit();
+        Optional<List<List<Commit>>> commits = findRootCommit(group, delivery).map(prevCommit -> commitsTillRoot(deliveryCommit, prevCommit));
```
JWGmeligMeyling: This is unfortunately very intensive as it has to query for the commit parents for every commit it traverses. For SKT this usually does not exceed 200 commits and thus 200 queries (with 20ms round trip that would take 4 seconds), but for bigger repositories this could potentially hang the Devhub server, which was also the reason why I limited the traversal that looks for unbuilt commits to 20 in the [`HooksResource`](https://github.com/devhub-tud/devhub/blob/master/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/HooksResource.java#L189-L197).

There are a couple of things we can do here:

1) Ignore the issue (I'd rather not).
2) Use [`blaze-persistence`](https://github.com/Blazebit/blaze-persistence/blob/master/documentation/src/main/asciidoc/core/manual/en_US/15_cte.adoc) so that we can create recursive CTE criteria queries for Hibernate. Pros: great framework, cons: even though we limit the `n` queries back to `1` query, Postgres is still not the fastest at processing these queries. Plus we would have to host a snapshot of `blaze-persistence` as it's currently not in Maven Central and under alpha release.
3) Use a Postgres CTE that selects the commit ids as Hibernate native query, and then find the commits for these ids. Cons: same query performance, error prone native SQL query in code base
4) Implement `git log` on the Git server (or see if we can use the existing branch endpoints for it) and use that.

src/main/java/nl/tudelft/ewi/devhub/server/backend/ConsecutiveBuildFailureBackend.java
```diff
@@ -0,0 +1,118 @@
...
+                .collect(toList());
+
+        final int index = assignments.indexOf(delivery.getAssignment());
+        final boolean firstAssignment = index < 1;
+
+        if (firstAssignment) {
+            return commits.firstCommitInRepo(group.getRepository());
+        } else {
+            final Assignment previousAssignment = assignments.get(index - 1);
+            return deliveries.getLastDelivery(previousAssignment, group).map(Delivery::getCommit);
```
JWGmeligMeyling: I think the commit for a delivery is nullable, so we should handle that case.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Delivery.java
```diff
@@ -239,6 +234,28 @@ public boolean isLate() {
         return dueDate != null && getTimestamp().after(dueDate);
     }
 
+    @JsonIgnore
+    public void setBuildFailed() {
+        Optional<Mastery> buildFailureMastery = findMasteryByDescription("Build for submitted commit fails");
```
JWGmeligMeyling: My original thought was to create `Mastery` subtypes (or `AutogradedMastery` subtypes) for this purpose, but I can live with this work around for now.

src/main/java/nl/tudelft/ewi/devhub/server/backend/ConsecutiveBuildFailureBackend.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+
+    protected static boolean hasFourFailingCommits(List<Commit> commits) {
+        int consecutive = 0;
+
+        for (Commit commit : commits) {
+            if (commit.getBuildResult().hasFailed()) {
+                consecutive++;
+                if (consecutive == MAX_CONSECUTIVE_BUILD_FAILURES) {
```
JWGmeligMeyling: This traversal treats merge commits invalidly. A failing merge commit for which the first child commit actually succeeded will break this chain. We have to treat this chain as a graph during traversal and look for any failing child commit instead, which is why I would look into the option of traversing generated `SuccessiveBuildFailureWarnings`.

src/main/java/nl/tudelft/ewi/devhub/server/backend/ConsecutiveBuildFailureBackend.java
```diff
@@ -0,0 +1,118 @@
...
+                }
+            } else {
+                consecutive = 0;
+            }
+        }
+
+        return false;
+    }
+
+    protected Optional<Commit> findRootCommit(Group group, Delivery delivery) {
```
JWGmeligMeyling: Isn't the `Group` already contained within the `Delivery`?

PR #421, opened 2017-04-05
Update pom.xml

ManInTheGitHub:
Integrate owasp dependency checker and update versions of jsch, hibernate validator and logback accordingly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
 
 		<guice.version>4.0</guice.version>
 		<hibernate.version>4.3.8.Final</hibernate.version>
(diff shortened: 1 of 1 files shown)


JWGmeligMeyling:
Resteasy been running on an ancient Jackson version for a while (2.7.x. as opposed to 2.9.x). With upgrading Jackson we have to upgrade Resteasy too, and the compatibility is not easy to figure out (you would have to look into the pom for resteasy-jackson2). I am not exactly sure what Jackson modules we use in Github (I've used JDK8 and Hibernate extensions sometimes, but I don't think we use them here).

PR #423, opened 2017-04-09
Move existing rubrics to the next course edition

### LINKED ISSUE #420: Migrate rubrics from previous course edition
When creating a new course edition it would be nice to have the rubrics of the previous course present

LiamClark:
If a course already had a previous edition we add the same rubrics
to the next edition.

Give the assignments an id to solve the ID generation deadlock.
Persist the course edition first then add the assignments.

The courseEdition-create-error was being referenced from a controller,
but not actually present in the translation file.

Also pulled in assertJ for writing tests.

fixes #420

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,6 +31,7 @@
 		<selenium.version>2.53.0</selenium.version>
 		<build.server.version>1.2.0</build.server.version>
 		<git.server.version>1.0.10</git.server.version>
+		<assertj.version>3.6.2</assertj.version>
 	</properties>
 
     <repositories>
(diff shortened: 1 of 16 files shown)


pom.xml
```diff
@@ -383,6 +384,12 @@
 			<version>1.4.1</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.assertj</groupId>
+			<artifactId>assertj-core</artifactId>
+			<version>${assertj.version}</version>
```
JWGmeligMeyling: Should be in `test` scope :wink:

Fastjur: [It appears to be](https://github.com/devhub-tud/devhub/pull/423/files/dacec5b3708936ff3b79ea44d5fc5a99b799b39d..e9f8e2091d22f38eed830e6f11c896c419fd0552#diff-600376dffeb79835ede4a0b285078036R391)

JWGmeligMeyling: Yes because it was added afterwards 😉

Issue #412, opened 2016-06-22
When session expires after 15 minutes, the grading form does not forward to the login page


JWGmeligMeyling:
When session expires after 15 minutes, the grading form does not forward to the login page. As the form fails to persist under the hood, but does not indicate the user, this can lead to unpreferable results.

PR #426, opened 2017-04-09
Partial solution for #412

JWGmeligMeyling:
Partial fix for #412 by redirecting the user to the login page if the update request fails.
However, due to adblock limitations, this approach is not guaranteed to work.
A proper solution should exist of creating an login API endpoint and a JS login dialog.


src/main/resources/static/js/assignment-rubrics.js
```diff
@@ -83,6 +83,13 @@ module.controller('StatisticsControl', function($scope, $http, $q) {
         if (assignment) {
             $http.put('json', assignment).then(function(res) {
                 $scope.assignment = res.data;
+            }, function() {
+                try {
+	                  location.reload();
```
Fastjur: This does not work in all browsers right? Or adblocked browsers

PR #427, opened 2017-04-15
Fix the webtests by upgrading Selenium

TimvdLippe:
The webtests were not running anymore, as the Selenium drivers were not compatible with the latest versions of Firefox. Therefore, the fix is to upgrade Selenium to 3.x and use the Chrome driver instead of the Firefox driver.

I have added build instructions to the `CONTRIBUTING.md` and migrated the webtests code. Note that we can not rely on Selenium 3.3.1, as that has a dependency on `Guava:21.0` which is incompatible with most of the other code (other devhub dependencies as well as querydsl).

Let's see if this thing passes on CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: java
 sudo: required
 dist: trusty
+# cache the dependency-check-maven plugin catalog (280MB+)
+cache:
+  directories: $HOME/.m2/repository/org/owasp/dependency-check-data/
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,13 @@ addons:
   apt:
     packages:
       - oracle-java8-installer
+      - google-chrome-stable
+    sources:
+      - google-chrome
 jdk:
   - oraclejdk8
 before_install:
+  - "which chromedriver"
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ addons:
 jdk:
   - oraclejdk8
 before_install:
+  - sudo apt-get install chromium-chromedriver
+before_script:
+  - "export PATH=$PATH:/usr/lib/chromium-browser/"
+  - "which chromedriver"
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ addons:
 jdk:
   - oraclejdk8
 before_install:
-  - "which chromedriver"
   - "export DISPLAY=:99.0"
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,12 +1,23 @@
...
+      - google-chrome-stable
+    sources:
+      - google-chrome
 jdk:
   - oraclejdk8
 before_install:
+  - sudo apt-get install chromium-chromedriver
+before_script:
+  - "export PATH=$PATH:/usr/lib/chromium-browser/"
+  - "which chromedriver"
```
LiamClark: We can remove this now, I mainly wanted this as a fall back for when the test might fail. and check if the driver plugin was properly installed

CONTRIBUTING.md
```diff
@@ -43,7 +43,14 @@ To successfully run and test Devhub, a couple of steps should be followed to set
 		1. Install "Lombok Plugin"
 		
 When the JPA plugin for Eclipse generates a lot of false errors, you can disable the validation under `Preferences > Validation > JPA Validator`.
-		
+
+### Running integration tests
```
LiamClark: Let's just call this webtests?

TimvdLippe:
@LiamClark done!

.travis.yml
```diff
@@ -1,12 +1,22 @@
 language: java
-sudo: false
+sudo: required
```
JWGmeligMeyling: So this means we're back at the old and non-dockerized Travis environment, does that have any long term consequences?

.travis.yml
```diff
@@ -1,12 +1,22 @@
...
+dist: trusty
+# cache the dependency-check-maven plugin catalog (280MB+)
+cache:
+  directories: $HOME/.m2/repository/org/owasp/dependency-check-data/
```
JWGmeligMeyling: Why not cache entire `.m2`?

.travis.yml
TimvdLippe: Um, not as far as I know.
image

.travis.yml
TimvdLippe: Could do that too indeed

.travis.yml
JWGmeligMeyling: Ah yes so it is the old and non-dockerized travis environment (hence already the higher boot time), but they do not intend shutting that down any time soon.

.travis.yml
JWGmeligMeyling: Allright! 👍

.travis.yml
Fastjur: I don't think the higher boot times have a marginal effect on the already minute long travis run anyway ;)

LiamClark: can't we use this somehow? https://docs.travis-ci.com/user/migrating-from-legacy/#How-do-I-install-APT-sources-and-packages%3F

JWGmeligMeyling: Lets not go through that pain :wink:

PR #432, opened 2017-04-16
Upgrade Mockito to 2.8.3

TimvdLippe:
This PR upgrades Mockito to 2.8.3. It migrates deprecated method calls, removes usage of the MockitoJUnitRule and removes unused stubbings as caught by strict stubbing.

### DIFF
--- a/src/test/java/nl/tudelft/ewi/devhub/modules/MockedGitoliteGitServerModule.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/modules/MockedGitoliteGitServerModule.java
@@ -126,7 +126,6 @@ public class MockedGitoliteGitServerModule extends AbstractModule {
 	protected void createMockedRepositoriesFolder() {
 		String repositoriesPath = repositoriesFolder.toPath().toString() + "/";
 		when(configuration.getGitoliteBaseUrl()).thenReturn(repositoriesPath);
-		when(configuration.getRepositoriesDirectory()).thenReturn(repositoriesFolder);
 		log.info("Initialized bare repository folder in {}", repositoriesPath);
 	}
 
--- a/src/test/java/nl/tudelft/ewi/devhub/modules/MockedGitoliteGitServerModule.java
+++ b/src/test/java/nl/tudelft/ewi/devhub/modules/MockedGitoliteGitServerModule.java
@@ -126,6 +126,7 @@ public class MockedGitoliteGitServerModule extends AbstractModule {
 	protected void createMockedRepositoriesFolder() {
 		String repositoriesPath = repositoriesFolder.toPath().toString() + "/";
 		when(configuration.getGitoliteBaseUrl()).thenReturn(repositoriesPath);
+        when(configuration.getRepositoriesDirectory()).thenReturn(repositoriesFolder);
 		log.info("Initialized bare repository folder in {}", repositoriesPath);
 	}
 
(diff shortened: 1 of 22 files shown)


TimvdLippe:
Hm, it seems to be breaking on the Mockito installation. It resolves the artifact to https://devhub.ewi.tudelft.nl/nexus/content/repositories/releases/org/mockito/mockito-core/2.8.3/mockito-core-2.8.3.pom which is no maven central. Running locally works just fine. Unsure what I can do here?

JWGmeligMeyling:
https://devhub.ewi.tudelft.nl/ falls back to maven central. Based on maven-repository.com, the latest version available on central is [`2.7.22`](https://mvnrepository.com/artifact/org.mockito/mockito-core). So it seems you're testing against a local snapshot after all? Or perhaps the artifact moved to another group id?

src/test/java/nl/tudelft/ewi/devhub/modules/MockedGitoliteGitServerModule.java
```diff
@@ -126,7 +126,6 @@ protected void createMockedMirrorsFolder() {
 	protected void createMockedRepositoriesFolder() {
 		String repositoriesPath = repositoriesFolder.toPath().toString() + "/";
 		when(configuration.getGitoliteBaseUrl()).thenReturn(repositoriesPath);
-		when(configuration.getRepositoriesDirectory()).thenReturn(repositoriesFolder);
```
JWGmeligMeyling: As far as I remember, this stubbing is required for testing repository removal. It may be unused now, but this stubbing is required for the mocked git server module to work as expected as far as I remember. I remember it being used in the version for the `git-server` tests (of which this one is a fork).

TimvdLippe: Leaving it in now will break the tests for strict stubbing. We can use the silent version if you want to? Still, I think it can be removed for now?

JWGmeligMeyling: Aha hm that's an interesting question. Because this module was never designed for mocking unit tests, it is designed to act as a git server accessed through a git client by binding some of the git server implementation to its client interfaces. I see why Mockito discourages strict stubbing for unit tests, but really Mockito was used here primarily to:

1) Allow tests to spy behaviour on some of the mocked behaviour
2) Not have to go through extending every single interface in the hierarchy ourself

Perhaps it's better to rewrite this module so that it doesn't use Mockito anymore? Because I can imagine future tests that use this module failing because not all stubs will be hit in every single tests (as obviously not every possible git server interaction is used in every test). What do you think?

TimvdLippe:
We can just use the silent version of the runner :) Seems like an
appropriate option here

src/test/java/nl/tudelft/ewi/devhub/modules/MockedGitoliteGitServerModule.java
TimvdLippe: I have reverted it now. For some reason the tests were not failing when I ran them locally, so let's see if CI complains.

JWGmeligMeyling:
Perhaps because this test is only run on `mvn integration-test`?

TimvdLippe:
Yeah, already pushed an update. Should pass soonTM

PR #440, opened 2017-04-25
Newly registerd users are getting a npe when logging in

LiamClark:
due to the group list being null by refreshing them when we persist,
we will get an empty list.


LiamClark:
This fixed the problem in production, merging this.

PR #450, opened 2017-05-05
Escape delivery ids in the group assignment view

LiamClark:
Id's above 1000 would get a , inserted by freemarker.
causing a 404 because we can't find the resource. the ?c
tells freemarker the number is for machine purposes and not for humans.
This escapes it from the freemarker number parsing.


JWGmeligMeyling:
Thanks!

PR #467, opened 2017-05-31
Abillity to collapse rubrics when grading

### LINKED ISSUE #460: Make position of Remarks text box absolute to screen
Right now, if I'm at the bottom of the page and I want to write a remark, I have to go all the way to the top of the page, before having to go down again. Might be easier if the Remarks text box just stayed at the bottom or the top of the page and scrolled down with you

LiamClark:
The current implementation has some performance problems on collapsing the table.

fixes #460

### DIFF
--- a/src/main/resources/templates/courses/assignments/group-delivery-review.ftl
+++ b/src/main/resources/templates/courses/assignments/group-delivery-review.ftl
@@ -80,12 +80,12 @@
 								[#list assignment.getTasks() as task]
 								<tbody class="task" data-id="${task.id?c}">
 									<tr class="active">
-										<td colspan="2"><strong>${task.description}</strong></td>
+										<td colspan="2"><span><strong>${task.description}<p class="octicon octicon-chevron-down" data-toggle="collapse" data-target=".collapse-target-${task?index}"></p></strong></span></td>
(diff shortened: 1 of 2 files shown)


JWGmeligMeyling:
Yup the performance issues are weird. Liam is currently profiling that. It could be that the set of elements is much larger than thought, or that the event handler is invoked multiple times. But it seems the code is just fine and that the problem is actually caused by the fact that collapse is generally slow in Webkit. In the latter case, having this feature not performing is still better than not having this feature at all I suppose...

src/main/resources/templates/courses/assignments/group-delivery-review.ftl
```diff
@@ -170,6 +170,20 @@
 	$(computeTotals)
 	$('[type="radio"]').on('change', computeTotals)
     $('[type="radio"]').on('change', persistMasteries)
+
+	var down = "octicon-chevron-down"
+	var right ="octicon-chevron-right"
+
+	$('[data-target*="collapse-target"]').click(function() {
```
TimvdLippe: I think this query-selector is rather expensive. It might be better to instead add a class `collapse-target` and do `document.querySelector('.collapse-target')` and then put the actual value in `data-target="${task?index}"`. You can then retrieve the index with `tableRow.dataset.target`

PR #468, opened 2017-06-01
Grading progress indicator

JWGmeligMeyling:
This pull request implements a progress bar indication that shows which % of characteristics has been filled in during grading.

screen shot 2017-06-01 at 11 08 38

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Assignment.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Assignment.java
@@ -8,6 +8,7 @@ import lombok.NoArgsConstructor;
 import lombok.ToString;
 import nl.tudelft.ewi.devhub.server.database.Base;
 import nl.tudelft.ewi.devhub.server.database.entities.identity.FKSegmentedIdentifierGenerator;
+import nl.tudelft.ewi.devhub.server.database.entities.rubrics.Characteristic;
 import nl.tudelft.ewi.devhub.server.database.entities.rubrics.DutchGradingStrategy;
 import nl.tudelft.ewi.devhub.server.database.entities.rubrics.GradingStrategy;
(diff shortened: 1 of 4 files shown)


Fastjur:
Also. Does this show the progress per group per excercise? Or the overall progress of 'one assignment'?

To me it looks like it is the first one, but I thought the idea was to implement the second one to see how much of one assignment overall has been reviewed.

AZWN:
@Fastjur This button is generated by the multiselect submodule plugin. Did you initialize your submodules?

(See also [#415](https://github.com/devhub-tud/devhub/pull/415#issuecomment-229866343))

Fastjur:
@AZWN Updating my submodules fixed the issue. I'm still gonna leave it on request changes for now until we have decided on the unreadable white text issue.

Thanks!

JWGmeligMeyling:
@Fastjur It shows the percentage of characteristics that have been filled in by a TA for a particular submission / Delivery. So the assign 100% means fully graded and that the TA is done. We could hide the label if the percentage is 0%, but the problem is probably difficult to solve for 0-~15%. I wouldn't hide the progress bar altogether as the difference in alignment may be confusing.

pilmus:
It looks different on my system (less nice imho). Possibly something to do with the code in the .less file?

afbeelding

sranjha:
Looks different here as well, this is what happens when the progress bar is full:
image

JWGmeligMeyling:
You have to compile the less stylesheet file first (usually included in compilation if you do this through Maven instead of IntelliJ): try running `mvn lesscss:compile`. If you want to do this through IntelliJ: hover the button in the lowerleft corner, click "Maven projects", hit execute Maven goal (fifth button in the toolbar) and type `lesscss:compile`. Probably its included in `Generate Sources and Update Folders" too but I am not sure.

Fastjur:
Right. The readability issue is a tough one. One could argue that the percentage isn't really that important until above 50% or more. I'll approve the changes

PR #469, opened 2017-06-01
Compute state and grade on review if rubrics and grading strategy is used

JWGmeligMeyling:
This PR disables the grade and submission state inputs for TA's whenever a grading strategy is in place combined with any rubrics. Instead these fields are now generated right after finishing the review.

We currently do not update the grade when there is any change to the rubrics (for example the weights, or adding a new rubric). When the grades are released (or re-released), the grades will be re-computed. We may want to change the workflow here as this may be surprising. Recomputing the existing grades and statusses whenever the rubrics are changed probably makes more sense, and would make our "release and compute all" logic unnecessary.

### DIFF
--- a/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/ProjectAssignmentsResource.java
+++ b/src/main/java/nl/tudelft/ewi/devhub/server/web/resources/ProjectAssignmentsResource.java
@@ -23,6 +23,8 @@ import nl.tudelft.ewi.devhub.server.database.entities.Group;
 import nl.tudelft.ewi.devhub.server.database.entities.RepositoryEntity;
 import nl.tudelft.ewi.devhub.server.database.entities.User;
 import nl.tudelft.ewi.devhub.server.database.entities.rubrics.Characteristic;
+import nl.tudelft.ewi.devhub.server.database.entities.rubrics.GradingException;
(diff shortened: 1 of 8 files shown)


src/main/java/nl/tudelft/ewi/devhub/server/web/resources/ProjectAssignmentsResource.java
```diff
@@ -340,14 +342,28 @@ public Response processReview(@Context HttpServletRequest request,
...
-        Double gradeValue = grade.isEmpty() ? null : Double.valueOf(grade);
         Delivery delivery = deliveries.find(group, deliveryId);
+        Assignment assignment = delivery.getAssignment();
+        GradingStrategy gradingStrategy = assignment.getGradingStrategy();
+
+        Double gradeValue;
+
+        if (assignment.isAssignmentHasRubrics()) {
+            // MissingRubricExceptions should not happen here as the review button
+            // is disabled until all characteristics have been filled in
```
Fastjur: Are we fully gonna rely on client side checking for this?

JWGmeligMeyling: No, the server actually throws a `MissingRubricException` but I didn't go through the effort of implementing exception handling for that because it simply cannot occur through Devhub (hence the comment)

PR #471, opened 2017-06-07
Store line change nr

herfstregen:


pom.xml
```diff
@@ -30,7 +30,7 @@
 		<logback.version>1.1.11</logback.version>
 		<selenium.version>3.0.0</selenium.version>
 		<build.server.version>1.2.0</build.server.version>
-		<git.server.version>1.0.10</git.server.version>
+		<git.server.version>1.0.11-SNAPSHOT</git.server.version>
```
Fastjur: I think this is causing your Travis CI error. Look at the error message:

`[ERROR] Failed to execute goal on project devhub: Could not resolve dependencies for project nl.devhub:devhub:jar:1.0-SNAPSHOT: Failed to collect dependencies at nl.tudelft.ewi.git:git-models:jar:1.0.11-SNAPSHOT: Failed to read artifact descriptor for nl.tudelft.ewi.git:git-models:jar:1.0.11-SNAPSHOT: Could not transfer artifact nl.tudelft.ewi.git:git-models:pom:1.0.11-SNAPSHOT from/to codehaus-snapshots (https://nexus.codehaus.org/snapshots/): nexus.codehaus.org: Unknown host nexus.codehaus.org -> [Help 1]`

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Commits.java
```diff
@@ -47,13 +48,17 @@ public Commits(final EntityManager entityManager, final RepositoriesApi reposito
...
 			commit.setCommitId(commitId);
 			commit.setRepository(repositoryEntity);
 			commit.setComments(Lists.newArrayList());
 			commit.setPushTime(new Date());
+			commit.setLinesAdded(diffModel.getLinesAdded());
+			commit.setLinesRemoved(diffModel.getLinesRemoved());
 
 			enhanceCommitSafely(repositoryEntity, commitId, commit);
 
+			//HERE IT IS SAVED
```
Fastjur: We know ;)

src/test/java/nl/tudelft/ewi/devhub/server/database/controllers/CommitsTest.java
```diff
@@ -101,6 +108,17 @@ public void testPersistWithParent() {
...
+
+	@Test
+	public void testCommitLineChanges() {
+
+		Commit commit = createCommit(group.getRepository());
+		assertTrue(commit.getLinesAdded() == 0);
+		commit.setLinesAdded(69);
+		commit.setLinesRemoved(666);
+		assertTrue(commit.getLinesAdded() == 69);
+		assertTrue(commit.getLinesRemoved() == 666);
```
Fastjur: This doesn't seem productive to me. All we are doing here is testing our setters and getters.

You should implement a test that tests if the amount of lines changed corresponds to some git diff.

Also, `assertTrue` does not produce a clear error message. Try to use `assertThat` or `assertEquals`

pom.xml
JWGmeligMeyling: Yup I am aware of it. But this branch needs git-server changes. I'll review those first and we'll release a new version. CI should pass afterwards.

Fastjur: Alright. I'll wait for the update.

I'm keeping it on request changes due to the semi-useless test in `CommitsTest.java`

src/main/java/nl/tudelft/ewi/devhub/server/database/controllers/Commits.java
asylunatic: I did not know, and I was really proud to find out ;p

src/test/java/nl/tudelft/ewi/devhub/server/database/controllers/CommitsTest.java
asylunatic: Well I wasn't sure it was productive either and actually suspected that. But inspecting the other tests, it seemed like the procedure the rest of the tests were following and @JWGmeligMeyling confirmed that it was a fine way of testing?

JWGmeligMeyling: These tests verify the integration with Hibernate. There are a couple of things that can go wrong with mapping the value:

1) The field is not properly mapped or not present in the `changelog.xml`: the test will not run at all and fail during the `EntityManagerFactory` initialization. This does not to be tested if there is any test in the test codebase that boots up an `EntityManagerFactory` (which all of these controller integration tests do).

2) The field is not mapped at all. In order to test these, one should override the default value (most likely `null`), flush the changes (so they are written to the database) and refresh the entity (so that the value from the database overrides the in-memory value if different) and assert that the value equals the expected value.

3) The field has unreasonable constaints on it. For example a foreign-key constraint to the wrong entity, or a size that is too small to fit its regular content. By using sane values for the test for 2, this can easily be covered in that test.

 In order to make the test complete, you should flush the changes to the database ([`entityManager.flush()`](http://docs.oracle.com/javaee/7/api/javax/persistence/EntityManager.html#flush--)), then refresh the entity ([`entityManager.refresh(commit)`](http://docs.oracle.com/javaee/7/api/javax/persistence/EntityManager.html#refresh-java.lang.Object-)) and finally assert that the values are correct.

asylunatic: We have adjusted the test in the updated branch (showLinesChanged). We've opened another pull request for that branch and are closing this request 😸

asylunatic:
We have adjusted the test in the updated branch (showLinesChanged). We've opened another pull request for that branch and are closing this request 😸

src/test/java/nl/tudelft/ewi/devhub/server/database/controllers/CommitsTest.java
JWGmeligMeyling: Test is adjusted fine there. But please continue the discussion on the above :wink:

PR #479, opened 2017-06-08
Notifications database (WIP), fixes #394 , fixes #324

### LINKED ISSUE #324: Indicate unread notifications (for example messages)

### LINKED ISSUE #394: Repository notifications for new comments, course wide notifications for TA's

ArjanSeijs:
- Added a database for notifications
- Added a database to assign notifications to multiple users with a hasRead field.
- Added a page where the notifications can be shown + a indicator when there are unread notifications.
- Notifications can be mark as read.
- Todo: The notifications need to be displayed on the notifications page.


ArjanSeijs:
@TimvdLippe It was a placeholder icon, we already changed it to a number indicator to show how many unread notifications it has. I already removed the image.

src/main/resources/templates/notifications.ftl
```diff
@@ -0,0 +1,43 @@
...
+    <table class="table table-bordered">
+        <tbody>
+[#if notifications??]
+    [#list notifications as notificationWithUser]
+        <tr>
+            <td>
+                <form action="${path}/markRead" method="post" class="pull-right">
+                    <input type="hidden" name="notificationId" value="${notificationWithUser.getNotification().getId()}">
+                    <button type="submit" class="btn btn-danger btn-sm" style="margin: 5px;">
+                        <i class="glyphicon glyphicon-remove-sign"></i> MARKREAD
```
pilmus: Might be nice to change this to MARK READ (so with the space?)

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/User.java
```diff
@@ -116,5 +114,23 @@ public boolean isPasswordMatch(String password) {
 		return password != null && this.password != null
 				&& BCrypt.checkpw(password, this.password);
 	}
+
+	public List<Notification> getNotifications() {
+		List<Notification> notifications = new ArrayList<Notification>();
```
JWGmeligMeyling: I think something like `Lists.transform(notificationsToUsersList, NotificationsToUser::getNotification)` could work too. No need to edit this.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/User.java
```diff
@@ -116,5 +114,23 @@ public boolean isPasswordMatch(String password) {
...
+		for (NotificationsToUsers notificationsToUsers: notificationsToUsersList) {
+			notifications.add(notificationsToUsers.getNotification());
+		}
+		return notifications;
+	}
+
+	public int unreadNotifications() {
+		int amount = 0;
```
JWGmeligMeyling: This can also be written as `notificationsToUserList.stream().filter(NotificationToUser::isRead).count()`.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/notifications/Notification.java
```diff
@@ -0,0 +1,51 @@
...
+    //Change this to Enum?
+    private String event;
+
+    @Column(name = "message")
+    @NotEmpty
+    private String message;
+
+    @Column(name = "link")
+    // Change this to URI later
+    private String link;
```
JWGmeligMeyling: Actually, change this to `URI` now and see what happens. We should discuss too whether we want to store the link with the notification, or refer to an viewable object instead. Not only is the domain entity now very aware on how its propagated through the DevHub UI, but this approach also fills the database with URI's that may not always remain accessible (if we have to rewrite our API for example). It is perhaps easier to reference to the `Issue`, `Commit`, `Comment` or whatever entity triggered the event and then use the URI inherited from that.

This could be implemented in various ways but its probably nicest to create an inheritance tree of `Notifications` such as `IssueNotification`, `CommitNotification`, etc. This is however not trivial to map in Hibernate (I would have to help with that). I am primarily trying to start a discussion here whether its the best idea to model it this way.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/notifications/Notification.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Id
+    @Column(name = "id")
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
+
+    @Column(name = "event")
+    @NotNull
+    //Change this to Enum?
+    private String event;
```
JWGmeligMeyling: Yes we should change this to an enum value, unless we use the inheritance tree as suggested below, in which case the event type can be deduced from its class.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/notifications/Notification.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Column(name = "message")
+    @NotEmpty
+    private String message;
+
+    @Column(name = "link")
+    // Change this to URI later
+    private String link;
+
+    @NotNull
```
JWGmeligMeyling: I am wondering whether we actually want this to be a `@NonNull` value? I can imagine some sort of system generated notifications to appear in the future?

src/main/resources/changelog.xml
```diff
@@ -1112,6 +1112,28 @@
...
+			<column name="link" type="varchar(255)"></column>
+			<column name="user_id" type="bigint"></column>
+			<column name="date" type="datetime"></column>
+		</createTable>
+		<createTable tableName="notifications_to_users">
+			<column name="user_id" type="bigint"></column>
+			<column name="notification_id" type="bigint"></column>
+			<column name="isRead" type="boolean"></column>
+		</createTable>
+		<!--<addPrimaryKey tableName="notification_to_users" columnNames="user_id, notification_id"/>-->
```
JWGmeligMeyling: I commented these constrains out because they didn't work yet, but we should fix them. Also, edit the changelog so that the newly added changelog is appended at the bottom (below Vera's changeset). This because the changelog will always be applied in-order.

src/main/resources/static/less/devhub.less
```diff
@@ -960,4 +960,21 @@ img.emoji {
...
 	margin-top: -3px !important;
+}
+p.notificationTitle {
+	font-size: x-large;
+}
+.hasNotification::after {
+	color: white;
+	background-color: red;
+	border-color: white;
+	content: attr(data-amount);
```
JWGmeligMeyling: Nice trick! 👍

src/test/java/nl/tudelft/ewi/devhub/webtests/NotificationTest.java
```diff
@@ -0,0 +1,72 @@
...
+        String url = getDriver().getCurrentUrl();
+
+        loginView.clickLoginButton();
+
+        loginView.waitUntilCurrentUrlDiffersFrom(url);
+
+        NotificationView notificationView  = new NotificationView(getDriver());
+        notificationView.invariant();
+
+        WebElement button = getDriver().findElement(MARK_READ_BUTTON);
```
JWGmeligMeyling: To be consistent with our Page Objects structure, the tests itself should not have to deal with XPath expressions or the webdriver. Instead this logic should be abstracted in the Page Objects itself. The `NotificationView` should expose an API such as: `notificationView().getUnreadNotifications().get(0).markRead()` and `notificationView().getUnreadNotifications().size()`.

src/test/java/nl/tudelft/ewi/devhub/webtests/views/NotificationView.java
```diff
@@ -0,0 +1,28 @@
...
+    private static final String NET_ID = "admin1";
+
+    public NotificationView(WebDriver driver) {
+        super(driver);
+    }
+
+    @Override
+    public void invariant() {
+        assertTrue("Wrong Path", currentPathEquals("/notifications/" + NET_ID));
+        assertNotNull(By.xpath(NOTIFICATION1));
```
JWGmeligMeyling: This is kind of dependent on the configuration on the bootstrapper, which in theory could be configurable per test. Please verify instead that the notification table is there, and expose a mechanism to list potential notifications (which are then asserted in the tests itself).

pilmus:
Could you guys add a way to remove notifications after you've read them?

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/notifications/Notification.java
ArjanSeijs: I think this can be changed to using an viewable object indeed.

Maybe it is possible to make an interface for this and then use that to store the viewable object?
Because unless we want to store extra information for notifications of different types and don't see the point of creating a separate class for each type.

In this case in the database there is still a field for the type (event) of the notification and a reference to the id of the viewable object.

JWGmeligMeyling: There already is such an interface, but its poorly named, `Base`. It requires implementations to implement a `getURI` method that returns the `URI` for its page within DevHub. Its implemented for example by 
[`AbstractIssue`](https://github.com/devhub-tud/devhub/blob/master/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/issues/AbstractIssue.java#L58), [`RepositoryEntity`](https://github.com/devhub-tud/devhub/blob/master/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/RepositoryEntity.java#L58) and [`Assignment`](https://github.com/devhub-tud/devhub/blob/master/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/Assignment.java#L60) so that users of these entities can easily find their respective page within DevHub. I remember helping @cashenkes to implement it in [`Comment`](https://github.com/devhub-tud/devhub/blob/ef4a592a47cd13c478b0ab1b030252d4ec8eeac6/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/comments/Comment.java#L36) too, so that clicking the comment notification would bring you to the right issue, pull request or commit (depending on whether it was a [`CommitComment`](https://github.com/devhub-tud/devhub/blob/master/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/comments/CommitComment.java) or an [`IssueComment`](https://github.com/devhub-tud/devhub/blob/master/src/main/java/nl/tudelft/ewi/devhub/server/database/entities/comments/IssueComment.java). I can't seem to find that piece of code back in the repository though.

Anyways, implementing this interface for your `Notification` class is not trivial. Because what would be the implementation of this `getURI` method? Implementing the interface does not solve the issue addressed above: that storing the URI at persist time is not a great idea, and the notification should instead be linked to the entity it references.

However, the entities it may reference are from different types (so far `Issues` and `PullRequests`, but presumably `Commits`, `Deliveries`, and `Comments` in the near future too). Not only are these different types, but they are stored in different tables as well. As there is no common ancestor (besides `java.lang.Object`) there is no way to map all these types to a single field. So if you want to have the option to reference any of these types, you find your self doing something like this:

```java
private Issue linkedIssue;
private PullRequest linkedPullRequest;
// Many others to come...

public URI getURI() {
    if (linkedIssue != null) return linkedIssue.getURI();
    if (linkedPullRequest != null) return linkedPullRequest.getURI();
    // Many others to come...
}
``` 

This is where you end up having a lot of fields which lead to an ambiguous API. What should happen if I for example do:

```java
Notification notification = new Notification();
notification.setPullRequest(pullRequestB);
notification.setLinkedIssue(issueA);
```

A user may expect that setting the linked issue overrides the previously set pull request. But as observed from the code for `getURI` above, we can see that if there is a linkedIssue, its URI will be returned, ignoring any value in the other fields.

So we don't want this. It would be preferable instead to have subtypes of `Notifications` (for example `PullRequestNotification` that references to a `PullRequest`.

By the way, keep in mind that if we take this route, there is no need to distinguish between `PullRequestNotifications` and `IssueNotifications` because in this special case, `PullRequest` and `Issue` share a queryable common ancestor: `AbstractIssue`.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/notifications/Notification.java
cashenkes: We can still use this field when we want to send a system generated message. Since we can set it to something like DevHubSystems for example to generate notifications like: "DevHubSystems requested changes on pull request ..."

JWGmeligMeyling: That could be the message indeed, but the field still has to be empty (as the type is User and there is no system user). Right?

cashenkes: Yeah that is true, I didnt think about that. Should we then maybe remove the @NonNull and check whether there is a user, and if there isnt one add it as a system notification? (or something similar)

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/notifications/Notification.java
ArjanSeijs: What I had in mind is:
```JAVA
public interface Viewable {
   public URI getUri()
}
```
This interface is then implemented by `PullRequest`, `Issue` and other classes for Notifications.
Then in `Notification` you get something like this:
```JAVA
@Column(name = "linkedViewable")
private Viewable linkedViewable

public URI getUri() {
    return Viewable.getUri();
}
```
And then in the database there is one field for the `ID` of the `Viewable` and one field for the `type` since different types can share the same ID. Is it possible to map this to different entities using hibernarnate?

When this is not the case there is still to possibility to use it in this way: 
```JAVA
@Column(name = "viewableId")
private int viewable;
@Colum(name = "type")
private Enum type

public URI getUri() {
    return getViewableObject().getUri();
}

public getViewableObject() {
    switch(type)
         case PULL_REQUEST:
                return query().from(pullRequests).where(pullRequest.id.equals(viewable));
         case ISSUE:
        // query for other viewable objects.
}
```
has the down side that viewable needs to be retrieved every time from the database the URI is requested. In this case using subtypes is probably a better idea.

src/main/resources/templates/notifications.ftl
```diff
@@ -0,0 +1,53 @@
...
+                    <button type="submit" class="btn btn-danger btn-sm" style="margin: 5px;">
+                        <i class="glyphicon glyphicon-remove-sign"></i> ${i18n.translate("block.my-notifications.mark-read")}
+                    </button>
+                </form>
+                [#else]
+                <span class="pull-right">
+                    <i class="pull-right">${i18n.translate("block.my-notifications.is-read")}</i>
+                </span>
+                [/#if]
+                <div>${notificationWithUser.getNotification().getSender().getName()}, ${notificationWithUser.getNotification().getEvent()} #${notificationWithUser.getNotification().getId()}</div>
```
sranjha: Alright, this may sound nitpicky but wouldn't it be an idea to display the issue id instead of the notification id? Displaying the latter seems rather arbitrary as it doesn't relate to the issue itself, whereas the issue id would be more relevant to the user as the notification would then clearly indicate the issue it's related to.

image
image

Like here it would be nice if the id displayed in the notification would correspond to the id of the actual issue.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/notifications/Notification.java
JWGmeligMeyling: Yup I think that would make sense.

src/main/resources/templates/notifications.ftl
JWGmeligMeyling: Agreed, having the ID of the notification there does not make any sense.  Especially as it is a non-meaningful key (just a global auto increment).

Fastjur:
This is starting to become a bit of a clusterfuck of a PR.

Perhaps splitting it up would be a good idea?

JWGmeligMeyling:
I have pushed some changes that change how we resolve the URI of the warning. I've also refactored how we create the notifications so it requires slightly less boilerplate.

src/main/resources/templates/notifications.ftl
```diff
@@ -0,0 +1,57 @@
...
+[#if notifications??]
+    [#list notifications as notificationWithUser]
+        <tr>
+            [#if notificationWithUser.isRead(user)]
+            <td class="notification read">
+            [#else]
+            <td class="notification unread">
+            [/#if]
+
+            <a href="${notificationWithUser.URI}">
```
JWGmeligMeyling: I think we should alter this form so that it points to an endpoint that both marks the notification as read and redirects to the underlying URI. We can then remove the form below.

src/main/java/nl/tudelft/ewi/devhub/server/database/entities/User.java
```diff
@@ -116,5 +119,26 @@ public boolean isPasswordMatch(String password) {
...
+    }
+
+    public List<Notification> getReadNotifications() {
+		return notifications.stream()
+				.filter(notification -> notification.getRecipients().get(User.this))
+				.collect(Collectors.toList());
+    }
+
+    public int unreadNotifications() {
+        return getUnreadNotificationsList().size();
```
JWGmeligMeyling: Check number of ran queries here.

src/main/java/nl/tudelft/ewi/devhub/server/web/resources/NotificationResource.java
```diff
@@ -0,0 +1,92 @@
...
+	@GET
+	@Path("{netId}")
+	public String showUserPage(@Context HttpServletRequest request, @PathParam("netId") String netId)
+			throws IOException, ApiError {
+
+
+		Map<String, Object> parameters = Maps.newHashMap();
+		parameters.put("user", currentUser);
+		parameters.put("path", request.getRequestURI());
+		parameters.put("notifications", currentUser.getNotifications());
```
JWGmeligMeyling: TODO: Limit to last `n` notifications.

